PR #39, opened 2018-12-06
add ruleset tests and fix padding bug

mitsuo0114:
Hi,

I have append Ruleset tests. and fixed typical Python 2 and 3 compatible bug of "/" operator in padding calculation.

https://www.python.org/dev/peps/pep-0238/


j4v:
Thanks @mitsuo0114, merged PR

PR #41, opened 2018-12-10
Feature/addrule tests

mitsuo0114:
Hi,

I have add some tests for ruleset.py.

Probably current fetching logic have some bugs because we can see following errors in travis output.
I think we should fix these errors.  But because the logic is complicated, I hope these tests should help us to proceed to fix at first.

```
Error: the file /home/travis/build/nccgroup/ScoutSuite/ScoutSuite/core/data/rulesets/default.json does not exist.
Error: the file /home/travis/build/nccgroup/ScoutSuite/ScoutSuite/core/data/rulesets/default.json does not exist.
```

### DIFF
--- a/tests/test-rules-ruleset.py
+++ b/tests/test-rules-ruleset.py
@@ -19,7 +19,6 @@ class TestAWSScout2RulesRuleset:
 
     def test_ruleset_class(self):
         test001 = Ruleset(filename=self.test_ruleset_001)
-
         assert (os.path.isdir(test001.rules_data_path))
         assert (os.path.isfile(test001.filename))
         assert (test001.name == "test-ruleset")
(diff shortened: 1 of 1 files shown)


j4v:
Thanks for this and you contributions to the project. The `core` error you posted is a bug from refactoring the code, I'll fix it ASAP and then merge in your PR.

mitsuo0114:
Thanks, after you have fix the error, I will append another assertion to confirm "printError" is not called  with patches in general cases.

j4v:
I've identified what's causing the errors, working on a fix.

j4v:
The errors are now fixed. The 3 remaining ones are expected.

j4v:
I've had to comment a number of assertions as they were causing tests to fail, could you please review.

Issue #60, opened 2018-12-18
UI light/dark theme


j4v:
In addition to updating bootstrap to V4 (https://github.com/nccgroup/ScoutSuite/issues/59), allow the ability to change between light/dark themes.

This should allow switching themes within a report.

j4v:
The https://github.com/nccgroup/ScoutSuite/issues/59 issue should take precedence over this one, as upgrading Bootstrap to v4 will be required to use up to date themes.

vifor2:
My bad, I should've went through the list of issues before starting, but since Iv'e already started working on this issue I think I'll complete it, make a pull request, merge it and then update it while doing #59. Here's a WIP for reference : 
![image](https://user-images.githubusercontent.com/17322874/51268580-7aee5180-198e-11e9-95a0-a7408db967e6.png)

j4v:
It was my mistake, I put https://github.com/nccgroup/ScoutSuite/issues/59 in Milestone 2 so logically you started with this.

Screenshot looks 💯, might just have to make a few changes when you upgrade BS.

Issue #59, opened 2018-12-14
UI improvement - Bootstrap version and highlights


j4v:
- Update Bootstrap to 4.1.3 (currently uses 3.0.3)
- Update jQuery to 3.3.1 (currently uses 1.11.2)

Use Bootstrap colors for titles (https://getbootstrap.com/docs/4.0/utilities/colors/). This is especially important as text-on-titles doesn't show very well:
![sc_2018-12-14_15h26m44s](https://user-images.githubusercontent.com/4206926/50008517-b9ba7180-ffb4-11e8-8898-a85966ad7868.jpg)

vifor2:
This is proving to be much longer/tedious than I intended, still WIP.

j4v:
Yep, it may not be as easy as it seems:
- Upgrading BS probably breaks the theme. If it's a valid solution you could swap the themes for pre-made light/dark themes (e.g. https://bootswatch.com/).
- In the current UI architecture, the label's class is used for drawing & logic. This is really messy and shouldn't be the case, so using proper BS success/warning/danger labels will probably involve decoupling these two elements.

vifor2:
Yes Iv'e already upgraded the themes, but haven't applied your second point yet, thank you for the suggestions 👍 .

The main issue with upgrading Bootstrap is that the current classes used for styling either no longer exist, have been renamed or have different properties, therefore it's what I am currently adjusting and it's starting to look more like what it is supposed to look like!

One annoyance that I haven't managed to fix **yet** is the collapsible sub-menu, on the current version on develop it expands towards the bottom of the navbar, while on my version it expands on the right. Still searching for a way to force it to expand towards the bottom without having to change the structure in the .html. 

Before expanding : 
![image](https://user-images.githubusercontent.com/17322874/51548208-e2484d80-1e35-11e9-9a10-201b739a4e9f.png)

After expanding : 
![image](https://user-images.githubusercontent.com/17322874/51548230-ef653c80-1e35-11e9-87cc-ce9d2134373b.png)

j4v:
@vifor2 has the "Bootstrap colors for titles" been implemented, or are you moving this to a separate issue?

PR #103, opened 2019-01-17
Improvement/60 light dark themes

vifor2:
This PR adds a secondary (dark) theme that can be toggled on by clicking on the slider after opening the settings' drop-down menu. The theme is applied instantly and reapplied if needed each time the .html report is opened.

The visual of both themes isn't definitive since @j4v brought to my attention that #59 should've been completed prior to #60. Therefore I'll make the required adjustments to #60 after completing #59, which includes new bootstrap 4 themes.

Here's a collage of the default theme (renamed to light) and dark theme :
image


ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
```diff
@@ -374,6 +382,35 @@ function toggleDetails(keyword, item) {
...
+    }
+};
+
+/**
+ * Sets the css file location received as the theme
+ * @param file
+ */
+function set_theme(file)
+{
+    var oldlink = document.getElementsByTagName("link").item(THEME_INDEX);
```
Aboisier: I think adding a tag to the link element holding the theme would be more reliable. This solution breaks if the lines are moved around or if we add another link tag before.

ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
```diff
@@ -374,6 +382,35 @@ function toggleDetails(keyword, item) {
...
+    oldlink.href = file;
+}
+
+/**
+ * Save the current theme on web storage
+ */
+window.onunload = function() {
+    localStorage.setItem("dark_theme", document.getElementById("dark_theme").checked);
```
Aboisier: I wonder if this is compliant with the Cookie Law.

> The Cookie Law is a piece of privacy legislation that requires websites to get consent from visitors to store or retrieve any information on a computer, smartphone or tablet.

j4v: I don't think that's an issue as it's not a website, just some local html being rendered in the browser.

ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
j4v: I think a light & dark themes are sufficient, but it would be nice if the implementation allows to easily swap themes.

vifor2:
> Not sure if this is a local issue but switching themes has no effect:
After a clean clone + execution I now see the issue you mention. When working on the project I modified files in /scoutsuite-report in order to skip the process of making a report each time, I must've poorly transfered over my changes to the /ScoutSuite. Will fix.

> There's also a `scoutsuite.css` file that gets dropped in the `scoutsuite-report` folder.
I'm confused as to why this happens, will look into it too.

ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
vifor2: Now using an index, thanks for the tip @Aboisier ! @j4v Currently the only work needed to swap out themes is swapping inc-bootstrap/css/bootstrap-dark.min.css for a different file of the right version (3 atm) and changing the value of the const DARK_THEME in scoutsuite.js accordingly.

Issue #10, opened 2018-09-11
Empty string passed to getElementById()


andresriancho:
![screenshot from 2018-09-11 14-00-19](https://user-images.githubusercontent.com/865200/45375345-1465bb00-b5cb-11e8-9cee-2567f64edabe.png)

Clicking on the `12` does nothing, and `Empty string passed to getElementById()` is shown in the browser JS console.

JS console line is: `Empty string passed to getElementById().       jquery-1.11.2.min.js:2:24473`

PR #105, opened 2019-01-17
Bugfix/10 missing ec2 network interfaces

Aboisier:
This fixes issue #10 . It seems like there was one too many loops, which made the UI iterate over the id's letters instead of over the resources. The badge was actually showing the length of the id.

image


ScoutSuite/output/data/html/partials/aws/services.cloudformation.regions.id.stacks.html
```diff
@@ -1,50 +1,52 @@
...
-    </script>
+<!-- Single CloudFormation stack template -->
+<script id="single_cloudformation_stack-template" type="text/x-handlebars-template">
+  <div style="text-align: right; padding-right: 10px; text-weight: bold;"><a href="javascript:hidePopup()">X</a></div>
+  {{> services.cloudformation.regions.id.stacks}}
+</script>
 
+<script>
+  var single_cloudformation_stack_template = Handlebars.compile($("#single_cloudformation_stack-template").html());
+</script>
```
Aboisier: I just somehow formatted this file, there are no changes except the indentation.

j4v:
💯

Aboisier:
> Seems good, hopefully there's less indentation changes in the next PRs!

Unfortunately, there's a lot of funky indentation in the html files. Perhaps we can do a PR just for fixing that, this way, we will avoid polluting future PRs.

PR #106, opened 2019-01-18
Fix Mutable Default argument value

mitsuo0114:
Hi,

Some functions in whole project do not align with best practice.
These may be cause of bug in the future. I have fixed these.

"Mutable Default Arguments"
https://docs.python-guide.org/writing/gotchas/


mitsuo0114:
Please let me know if I need to add more test for this PR.

The coverage ratio was decreased because Some functions are not covered by current unit tests but the number of lines are increased.

Aboisier:
I changed the target branch to develop. This is currently an undocumented convention, it should appear in the contributing.md in the next few days.

Aboisier:
Thanks for your contribution Mitsuo!

PR #109, opened 2019-01-18
Master to develop

Aboisier:
Master is currently ahead of develop. Let's try avoiding this. I suggest setting the default branch to `develop` instead of `master`. This way, new PRs will be against `develop` by default.

One argument against using develop as the default branch is that people looking at the repo will see the develop branch by default.


j4v:
Taking into account the later, wouldn't it be easier to simply draft a `contributions.md` file stating to push PRs to `develop` and reject PRs made directly to `master`?

Aboisier:
I think that's a good solution. I'm currently working on a `contributing.md` file, I'll make it clear that PRs should target `develop`.

PR #111, opened 2019-01-19
Bool to enabled helper

Aboisier:
This was initially implemented for a proprietary functionality, but I decided to refactor a couple of other pieces of code.

A couple of rules were using magic strings (`"Disabled"` and `"Enabled"` instead of booleans). I converted the strings to boolean, and added a helper to convert boolean values to strings.

### DIFF
--- a/ScoutSuite/providers/aws/services/s3.py
+++ b/ScoutSuite/providers/aws/services/s3.py
@@ -265,15 +265,15 @@ def get_s3_bucket_secure_transport(api_client, bucket_name, bucket_info):
 def get_s3_bucket_versioning(api_client, bucket_name, bucket_info):
     try:
         versioning = api_client.get_bucket_versioning(Bucket=bucket_name)
-        bucket_info['versioning_status_enabled'] =  _enabled_to_bool(versioning.get('Status'))
-        bucket_info['version_mfa_delete_enabled'] = _enabled_to_bool(versioning.get('MFADelete'))
+        bucket_info['versioning_status_enabled'] = _status_to_bool(versioning.get('Status'))
+        bucket_info['version_mfa_delete_enabled'] = _status_to_bool(versioning.get('MFADelete'))
         return True
     except Exception as e:
         bucket_info['versioning_status_enabled'] = None
         bucket_info['version_mfa_delete_enabled'] = None
         return False
 
-def _enabled_to_bool(value):
+def _status_to_bool(value):
     """ Converts a string to True if it is equal to 'Enabled' or to False otherwise. """
     return value == 'Enabled'
 
--- a/ScoutSuite/providers/aws/services/s3.py
+++ b/ScoutSuite/providers/aws/services/s3.py
@@ -293,7 +293,7 @@ def get_s3_bucket_logging(api_client, bucket_name, bucket_info):
 def get_s3_bucket_webhosting(api_client, bucket_name, bucket_info):
     try:
         result = api_client.get_bucket_website(Bucket=bucket_name)
-        bucket_info['web_hosting_enabled'] = IndexDocument' in result
+        bucket_info['web_hosting_enabled'] = 'IndexDocument' in result
         return True
     except Exception as e:
         # TODO: distinguish permission denied from  'NoSuchWebsiteConfiguration' errors
--- a/ScoutSuite/providers/aws/services/s3.py
+++ b/ScoutSuite/providers/aws/services/s3.py
@@ -265,13 +265,17 @@ def get_s3_bucket_secure_transport(api_client, bucket_name, bucket_info):
 def get_s3_bucket_versioning(api_client, bucket_name, bucket_info):
     try:
         versioning = api_client.get_bucket_versioning(Bucket=bucket_name)
-        bucket_info['versioning_status_enabled'] = 'Status' in versioning
-        bucket_info['version_mfa_delete_enabled'] = 'MFADelete' in versioning
+        bucket_info['versioning_status_enabled'] =  _enabled_to_bool(versioning.get('Status'))
+        bucket_info['version_mfa_delete_enabled'] = _enabled_to_bool(versioning.get('MFADelete'))
         return True
     except Exception as e:
         bucket_info['versioning_status_enabled'] = None
         bucket_info['version_mfa_delete_enabled'] = None
         return False
+
+def _enabled_to_bool(value):
+    """ Converts a string to True if it is equal to 'Enabled' or to False otherwise. """
+    return value == 'Enabled'
 
 
 def get_s3_bucket_logging(api_client, bucket_name, bucket_info):
--- a/ScoutSuite/providers/aws/services/s3.py
+++ b/ScoutSuite/providers/aws/services/s3.py
@@ -240,7 +240,7 @@ def get_s3_bucket_policy(api_client, bucket_name, bucket_info):
 def get_s3_bucket_secure_transport(api_client, bucket_name, bucket_info):
     try:
         if 'policy' in bucket_info:
-            bucket_info['secure_transport'] = 'Disabled'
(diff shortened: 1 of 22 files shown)


ScoutSuite/providers/gcp/services/cloudstorage.py
```diff
@@ -54,15 +53,11 @@ def parse_buckets(self, bucket, params):
 
 def get_cloudstorage_bucket_logging(bucket, bucket_dict):
     try:
-        logging = bucket.get_logging()
-        if logging:
-            bucket_dict['logging_status'] = 'Enabled'
-        else:
-            bucket_dict['logging_status'] = 'Disabled'
+        bucket_dict['logging_enabled'] = bucket.get_logging() != None
```
Remi05: Do we know for sure what the return type for `bucket.get_logging()` is? If it is anything else than a boolean, then the behaviour would change, as Python can treat various types as booleans implicitly. See the [following link](https://anh.cs.luc.edu/python/hands-on/3.1/handsonHtml/boolean.html) for more info.

Using `bucket_dict['logging_enabled'] = bool(bucket.get_logging())` instead would preserve the current behaviour.

ScoutSuite/providers/aws/services/s3.py
```diff
@@ -293,26 +293,26 @@ def get_s3_bucket_logging(api_client, bucket_name, bucket_info):
 def get_s3_bucket_webhosting(api_client, bucket_name, bucket_info):
     try:
         result = api_client.get_bucket_website(Bucket=bucket_name)
-        bucket_info['web_hosting'] = 'Enabled' if 'IndexDocument' in result else 'Disabled'
+        bucket_info['web_hosting_enabled'] = 'IndexDocument' in result
```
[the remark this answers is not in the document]
Aboisier: Not sure how the behaviour could be different here.  Do you have a specific case in mind?

ScoutSuite/providers/aws/services/s3.py
```diff
@@ -251,26 +251,26 @@ def get_s3_bucket_secure_transport(api_client, bucket_name, bucket_info):
...
         return False
 
 
 def get_s3_bucket_versioning(api_client, bucket_name, bucket_info):
     try:
         versioning = api_client.get_bucket_versioning(Bucket=bucket_name)
-        bucket_info['versioning_status'] = versioning['Status'] if 'Status' in versioning else 'Disabled'
-        bucket_info['version_mfa_delete'] = versioning['MFADelete'] if 'MFADelete' in versioning else 'Disabled'
+        bucket_info['versioning_status_enabled'] = 'Status' in versioning
+        bucket_info['version_mfa_delete_enabled'] = 'MFADelete' in versioning
```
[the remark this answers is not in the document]
Aboisier: True, there were so many cases to fix, I did it a bit too quickly!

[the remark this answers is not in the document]
Aboisier: Definitely. I'll fix this.

ScoutSuite/providers/gcp/services/cloudstorage.py
Aboisier: `get_logging` returns either a dictionary or `None`. See [this link](https://googleapis.github.io/google-cloud-python/latest/storage/buckets.html#google.cloud.storage.bucket.Bucket.get_logging).

ScoutSuite/providers/aws/services/s3.py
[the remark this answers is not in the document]
Aboisier: I believe this fixes it. It is also a bit more readable imho. What do you think?

j4v:
> This was initially implemented for a proprietary functionality, but I decided to refactor a couple of other pieces of code.
> 
> A couple of rules were using magic strings (`"Disabled"` and `"Enabled"` instead of booleans). I converted the strings to boolean, and added a helper to convert boolean values to strings.

I quite like this feature, although it might be a bit confusing as a large amount of rules/html partials will use the "old style" `Enabled`/`Disabled` strings. Just make sure you use this going forward, and hopefully we'll get around to updating existing rules.

ScoutSuite/providers/gcp/services/cloudsql.py
```diff
@@ -82,3 +76,9 @@ def _get_instance_backups(self, instance, params):
         except Exception as e:
             printError('Failed to parse backups for SQL instance %s: %s' % (instance['name'], e))
             return None
+
+    def _is_log_enabled(self, instance) :
+        return instance['settings']['backupConfiguration'].get('binaryLogEnabled')
+
+    def _is_ssl_required(self, instance):
+        return instance['settings']['ipConfiguration'].get('requireSsl')
```
Remi05: If `instance['settings']['ipConfiguration']` doesn't contain `'requireSsl'`, then `None` will be returned instead of `False`.

```suggestion
        return instance['settings']['ipConfiguration'].get('requireSsl', False)
```

ScoutSuite/providers/gcp/services/cloudsql.py
```diff
@@ -82,3 +76,9 @@ def _get_instance_backups(self, instance, params):
         except Exception as e:
             printError('Failed to parse backups for SQL instance %s: %s' % (instance['name'], e))
             return None
+
+    def _is_log_enabled(self, instance) :
+        return instance['settings']['backupConfiguration'].get('binaryLogEnabled')
```
Remi05: If `instance['settings']['backupConfiguration']` doesn't contain `'binaryLogEnabled'`, then `None` will be returned instead of `False`.

```suggestion
        return instance['settings']['backupConfiguration'].get('binaryLogEnabled', False)
```

ScoutSuite/providers/gcp/services/cloudstorage.py
Aboisier: The behavior is different but the outcome is still correct, according to the API. It specifically says `a dict with keys or None`.

As for the `!=`, I suggest we use GitHub's online file editing tool instead of leaving comments for small style-related issues, excepted if it's a recurring error.

ScoutSuite/providers/aws/services/s3.py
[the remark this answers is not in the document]
Remi05: Love the new approach!

ScoutSuite/providers/gcp/services/cloudsql.py
Remi05: See the [doc on `dictionary.get()`](https://www.tutorialspoint.com/python/dictionary_get.htm) for more info.

ScoutSuite/providers/gcp/services/cloudsql.py
Aboisier: According to the [api reference](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances), `binaryLogEnabled` should be defined. Do we want to assume it is disabled if it is undefined? This solution will appear as "Unknown" in the report. I'm not sure whether it will get flagged or not, I'll investigate.

Aboisier: Can confirm, gcp will return False if the config is disabled.

ScoutSuite/providers/gcp/services/cloudstorage.py
Remi05: Oh, I didn't see the "w/ keys" part! Also, agreed for the syntax fix!

ScoutSuite/providers/gcp/services/cloudsql.py
Remi05: As you seem fit, I simply noticed the behaviour changed since your previous implementation of this method!

Aboisier: Alright thanks!

Issue #81, opened 2018-12-20
Properly implement Azure authentication


j4v:
Properly implement authentication for the Azure provider. Current implementation in `providers/azure/provider.py` only supports service principals [in a limited manner].

Refs:
https://docs.microsoft.com/en-us/python/azure/python-sdk-azure-authenticate?view=azure-python
https://azure.microsoft.com/en-ca/resources/samples/data-lake-analytics-python-auth-options/
https://github.com/AzureAD/azure-activedirectory-library-for-python/wiki
https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-scenarios#application-types-and-scenarios
https://docs.microsoft.com/en-us/python/api/overview/azure/activedirectory?view=azure-python

Once this is done, update README to explain how to authenticate.

mitsuo0114:
Does this ticket intend to append "App Registration" authentication like tenable do?[1] or implement 
ManagementClients like providers/aws/provider.py and official sample code?[2]

Because I have development experience of Azure REST API with Service Principal, I may be able to contribute to this ticket.

[1] https://docs.tenable.com/other/azure/Content/Auditing.htm
[2] https://github.com/Azure-Samples/virtual-machines-python-manage/blob/master/example.py

zer0x64:
Implemented in #112, closing

j4v:
Testing this out I see that credentials are prompted interactively:

![sc_2019-02-01_10h04m56s](https://user-images.githubusercontent.com/4206926/52113168-e845e780-2608-11e9-8bf2-f98d2fac476c.jpg)

As for AWS/GCP, all CLI parameters should be passed directly. The rationale for this is that asking for credentials interactively doesn't allow to run Scout programatically.

One option would be to add a `--interactive` flag to *also* support how it's currently implemented, but we really need to be able to run Scout from code.

j4v:
@zer0x64 I think this can be closed now?

zer0x64:
Yes!

PR #112, opened 2019-01-20
Feature/81 Properly implement Azure authentication

zer0x64:
This pull request aims to fix issue #81 by properly implementing azure authentication in a similar way used for the other providers, using credentials provided by the provider's CLI tool. 

This way, every authentication scheme provided by azure-cli can be used to run ScoutSuite. This includes interactive login using OAuth and a browser, basic username/password authentication, and service principals credentials.


ScoutSuite/providers/azure/provider.py
```diff
@@ -2,29 +2,30 @@
...
     """
-    Implements provider for AWS
+    Implements provider for Azure
     """
 
     def __init__(self, project_id=None, organization_id=None,
                  report_dir=None, timestamp=None, services=[], skipped_services=[], thread_config=4, **kwargs):
 
-        self.profile = 'aws-profile'  # TODO this is aws-specific
+        self.profile = 'azure-profile'  # TODO this is aws-specific
```
Aboisier: Is that comment still relevant?

ScoutSuite/providers/azure/provider.py
```diff
@@ -37,37 +38,16 @@ def __init__(self, project_id=None, organization_id=None,
...
-
-            credentials = ServicePrincipalCredentials(
-                client_id=CLIENT,
-                secret=KEY,
-                tenant=TENANT_ID
-            )
-
-            self.credentials = AzureCredentials(credentials, SUBSCRIPTION_ID)
-
+            cli_credentials, self.aws_account_id = get_azure_cli_credentials()          #TODO: Remove aws_account_id
```
Aboisier: Can you complete that todo in this PR or is it out of the scope?

ScoutSuite/providers/azure/provider.py
zer0x64: Yes it is! self.profile and self.aws_account_id are members that are only really useful for AWS provider, but breaks if removed from other the other providers. Note that these are in GCP too.

self.profile is used in base/provider.py to generate report and has a comment saying that it would be a significant re-write to get rid of it, so I think that might be better to do a specific PR for this refactoring.

ScoutSuite/providers/azure/provider.py
zer0x64: As with the other issue, this one is used in __main__.py at line 132 and might lead to a refactoring rabbit hole. Again, I believe it would be better for it to have it's own PR and keep this one only for adding the new feature.

ScoutSuite/providers/azure/provider.py
Aboisier: Ok, no worries!

ScoutSuite/providers/azure/provider.py
Aboisier: Same here 👍

zer0x64:
> > azure-cli-core 2.0.55 has requirement pyyaml>=4.2b1, but you'll have pyyaml 3.13 which is incompatible.

Seems to be related to https://github.com/pypa/pip/issues/988. Nice catch! Will try the proposed solution.

README.md
```diff
@@ -173,4 +173,6 @@ To scan a GCP ...
 
 #### Azure
 
-TODO
+Using a computer already configured to use azure-cli, you may use Scout using the following command:
```
j4v: I don't know if this is possible, but with `azure-cli` can you have multiple accounts configured at once? If so, we'd need a way to tell Scout which configured account to use.

j4v:
Can you also update `ScoutSuite/cli_parser.py`:
```
  --provider {aws,gcp,azure}
                        The cloud provider to scan (currently supports AWS
                        ('aws') and GCP ('gcp')
```

README.md
zer0x64: Currently looking into the multiple account. This doesn't seem really documented, but the SDK code is open-source and there are comments regarding the arguments.

zer0x64: Hi! So for using multiple user accounts, this does not seems to be directly possible, as stated in https://github.com/Azure/azure-xplat-cli/issues/2198. 

However, it may be possible to switch between different subscription, as the azure-cli tool permits it.

j4v: Could you please look into https://ncc-scout.slack.com/archives/CDCSGNQ8M/p1548008131030700. While using the `azure-cli` seems to be very convenient, it might not be the way to go. One of the primary reasons why I believe ADAL is a better option is that with `azure-cli` it will be hard to run the tool programatically (e.g. spin up a container to run Scout, or run it in lambda, etc. - these are scenarios we are currently using).

It will require a bit more core to use ADAL as you'll have to handle all (4?) authentication scenarios, but it will be beneficial in the long run.

ScoutSuite/providers/azure/provider.py
```diff
@@ -35,39 +39,55 @@ def __init__(self, project_id=None, organization_id=None,
...
+
+                return True
+            elif azure_user_credentials:
+                username = input("Username: ")
+                password = getpass("Password: ")
+
+                credentials = UserPassCredentials(username, password)
+                self.aws_account_id = ""  # TODO this is for AWS
+                self.credentials = AzureCredentials(credentials, self.aws_account_id)
+                return True
```
Aboisier: I suggest we break down that method into multiple smaller methods. For example, `azure_cli_authentication`, `azure_msi_authentication`, `azure_file_auth`, etc. That would increase readability and 
 
It could be interesting to eventually implement a strategy pattern or something like that instead of that if/elif/elif/...

ScoutSuite/providers/azure/provider.py
```diff
@@ -35,39 +39,55 @@ def __init__(self, project_id=None, organization_id=None,
...
+            elif azure_msi:
+                # TODO test
+                credentials = MSIAuthentication()
+                self.aws_account_id = "aaa"
+                self.credentials = AzureCredentials(credentials, self.aws_account_id)
+                return True
+            elif azure_file_auth:
+                # TODO
+                return True
```
Aboisier: Do you plan on implementing this authentication method in this PR? If not, does the `return True` implies the connection was successful? If so, we should probably tell the user this is currently not implemented.

ScoutSuite/providers/azure/provider.py
```diff
@@ -35,39 +39,55 @@ def __init__(self, project_id=None, organization_id=None,
...
-            )
-
-            self.credentials = AzureCredentials(credentials, SUBSCRIPTION_ID)
-
-            return True
-
+            if azure_cli:
+                cli_credentials, self.aws_account_id = get_azure_cli_credentials()  # TODO: Remove aws_account_id
+                self.credentials = AzureCredentials(cli_credentials, self.aws_account_id)
+                print(cli_credentials)
```
Aboisier: Do we want to print the credentials?

ScoutSuite/providers/azure/provider.py
```diff
@@ -35,39 +39,55 @@ def __init__(self, project_id=None, organization_id=None,
...
+                return True
+            elif azure_msi:
+                # TODO test
+                credentials = MSIAuthentication()
+                self.aws_account_id = "aaa"
+                self.credentials = AzureCredentials(credentials, self.aws_account_id)
+                return True
```
Aboisier: If this is untested, I should we at least inform the user this functionality is experimental.

ScoutSuite/providers/azure/provider.py
zer0x64: Good idea, I'll do it!

ScoutSuite/providers/azure/provider.py
zer0x64: Yes, I plan on doing this on this PR, I juste haven't done it yet because of permission issue!

ScoutSuite/providers/azure/provider.py
zer0x64: lol. Forgot it. But at least it didn't print the credentials, only the type ;)

ScoutSuite/providers/azure/provider.py
zer0x64: Yep! I still haven't updated the docs! I just wanted to push what I did, I forgot to write that it wasn't ready to pull!

ScoutSuite/providers/azure/provider.py
zer0x64: Oops.

ScoutSuite/providers/azure/provider.py
zer0x64: Done!

ScoutSuite/providers/azure/provider.py
zer0x64: It is implemented now!

ScoutSuite/providers/azure/provider.py
zer0x64: Hi!

If you want, I could check this out while I refactor the cli parser, as this pattern would also be useful for the other provider. For now, at least it's in a working state and I'll clean it up for good on the next PR!

zer0x64:
The MSI authentication is currently untested. I don't know if the problem is with my azure setup or if I don't have the required permission to set it up properly. I will finish up the README when the microsoft doc is back up again(it is down because of certificate problems).

ScoutSuite/providers/azure/provider.py
j4v: > I could check this out while I refactor the cli parser, as this pattern would also be useful for the other provider.

Agreed, the CLI implementation can/should be improved as currently it's just a soup of parameters. The parameters should have a hierarchy so that it's clear to which provider they apply.

zer0x64:
MSI Auth now working and tested with Reader role. I think all that's left is to update README.md!

README.md
```diff
@@ -175,4 +205,21 @@ To scan a GCP ...
 
 Using a computer already configured to use azure-cli, you may use Scout using the following command:
 
-    $ python Scout.py --provider azure
+    $ python Scout.py --provider azure --azure-cli
+    
+When using Scout in an Azure virtual machine with the Reader role on it's identity, you may use
```
Aboisier: Should be its instead of it's.

```suggestion
When using Scout in an Azure virtual machine with the Reader role on its identity, you may use 
```

We could also say the following though

```suggestion
When using Scout in an Azure virtual machine with the Reader role, you may use 
```

README.md
```diff
@@ -175,4 +205,21 @@ To scan a GCP ...
 
 Using a computer already configured to use azure-cli, you may use Scout using the following command:
 
-    $ python Scout.py --provider azure
+    $ python Scout.py --provider azure --azure-cli
+    
+When using Scout in an Azure virtual machine with the Reader role on it's identity, you may use 
+Scout using the following command:
```
Aboisier: I think we need to refactor this sentence.

ScoutSuite/providers/azure/provider.py
Aboisier: Fair enough 👍

j4v:
Excellent work @zer0x64, really excited to start seeing Azure functionalities!

Issue #36, opened 2018-12-03
Opening tab with anchor link


j4v:
It isn't possible to open links in new tabs as the anchor isn't included (it always opens the `<report>.html` file). Implementing this would help with navigation.

e.g. https://stackoverflow.com/questions/15678511/opening-tab-with-anchor-link

Aboisier:
Implemented in #113 .

PR #113, opened 2019-01-21
Feature/36 open link with anchor

Aboisier:
With this PR, opening a link containing an anchor will open the right page instead of the main dashboard. Implements #36 .

example


j4v:
gangster

PR #115, opened 2019-01-23
Feature/59 bootstrap jquery upgrades

vifor2:
In this PR:

1. Boostrap was updated from 3.0.3 to 4.2.1:
    - In ScoutSuite/output/data/includes.zip/inc-bootstrap the main CSS file, the theme CSS files and the JavaScript file were replaced.
    - Many classes were renamed/dropped in Bootstrap 4 and therefore the classes of many elements have been updated accordingly.
    - Glyphicon is no longer supported for free on BS4 and therefore has been replaced by Font Awesome (added inc-fontawesome in includes.zip).

2. Jquery was updated from 1.11.2 to 3.3.1.

Will include pictures in the next post.

N.B. The sub-menu that appears under the main navbar is not done in a proper way, it is a temporary fix since it was blocking me. Also, I tried to stay within scope of the issue and therefore tried to make the least changes possible to the architecture of the UI, I plan on creating an issue so that @j4v's concern can be addressed eventually :

>In the current UI architecture, the label's class is used for drawing & logic. This is really messy and shouldn't be the case, so using proper BS success/warning/danger labels will probably involve decoupling these two elements.


vifor2:
Here's a video instead.
[ScoutSuite.zip](https://github.com/nccgroup/ScoutSuite/files/2787938/ScoutSuite.zip)

j4v:
A couple of points - some relating to the navbar so you might already have noticed.

The elements are no longer grayed out when they are unavailable:

sc_2019-01-23_19h10m12s

When you click a "disabled link", the URL is the html page (see link URL at the bottom):

sc_2019-01-23_19h13m02s

Any reason the middle column is smaller here?

sc_2019-01-23_19h17m09s

It isn't the case for other elements

Also is it possible to make the text a bit smaller in the dashboard? 

Other then that the look and feel are great, the above are just details.

vifor2:
Thank you for the observations @j4v! I'll try to get those points fixed after school and no, I have no idea why that column is smaller 😅

j4v:
> I have no idea why that column is smaller

I think it's because the content doesn't take all the space. I'd think it's a small fix, probably adding a BS table class.

One other small thing (which was there before this PR but would be nice to fix is that long elements spill out of the leftmost column:

sc_2019-01-24_09h38m26s

A quick fix for this could be to redact long elements? e.g. for this case to display `ApplicationAutoScalingForAmazonAppStre...`.

vifor2:
Updated PR with fixes suggested : 
- Disabled elements are back to being greyed out/unclickable.
- Fixed middle column size.
- Reduced font size of last run dashboard to match the version on develop.
- Text in IAM Policies now breaks instead of overflowing.

image
image
image

PR #116, opened 2019-01-23
Feature/12-proprietary

Aboisier:
This PR is part of a feature done on the proprietary side. This removes the <root-account> specific credential report page. We now have a credential reports page which contains the credential reports of each user.

### DIFF
--- a/ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
+++ b/ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
@@ -1039,10 +1039,6 @@ function get_resource_path(path) {
         resource_path_array = resource_path.split('.');
         last_value = resource_path_array.pop();
         resource_path = 'services.' + resource_path_array.join('.');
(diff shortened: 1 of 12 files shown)


ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
```diff
@@ -1039,10 +1038,6 @@ function get_resource_path(path) {
         resource_path_array = resource_path.split('.');
         last_value = resource_path_array.pop();
         resource_path = 'services.' + resource_path_array.join('.');
-        // Fix for issue #79
-        if (last_value == '<root_account>') {
-            resource_path += '.' + last_value;
-        };
```
Aboisier: This page does not exist anymore. Clicking on root-account specific rules will redirect to the credential-reports dashboard with the <root-account> focused.

j4v:
Great job, this will be a nice improvement.

PR #142, opened 2019-02-04
Release v4.1.0

Aboisier:
Release v4.1.0.

Please also see the [proprietary PR](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/66).


ScoutSuite/core/processingengine.py
```diff
@@ -65,3 +65,7 @@ def run(self, cloud_provider, skip_dashboard = False):
                     cloud_provider.services[service][self.ruleset.rule_type][rule.key]['checked_items'] = 0
                     cloud_provider.services[service][self.ruleset.rule_type][rule.key]['flagged_items'] = 0
 
+    def _filter_rules(self, rules, services):
+        return { rule: rules[rule] for rule in rules if rule.split('.')[0] in services }
```
zer0x64: Didn't see this pass, but isn't using "rule" and "rules" a litte confusing? I would rather use "r" as variable name for the single rules.

zer0x64: Also, it seems to be more pythonic to use items(), since direct indexing of dictionaries can cause a crash. Something like this:
return { rule: value for key, value in rules.items() if key.split('.')[0] in services }

Not actually sure what the key(rule) and values(rules[rule]) means with that specific data structure

j4v:
I would wait until the CLI has been refactored (i.e. until https://github.com/nccgroup/ScoutSuite/tree/refactor/127-cli-refactoring is merged) before making the 4.1.0 release.

Main reason is to that Azure authentication doesn't change too much from 4.1.0 to 4.2.0. Sounds good?

misg:
Tested it with Azure and GCP and everything looks fine. :100: 
Could someone test it with AWS?

j4v:
> Could someone test it with AWS?

Looks fine to me 🥇

PR #149, opened 2019-02-05
feature/#26-new-finding-azure-sql-no-auditing-nor-threat-detection

misg:
This PR adds support for SQL service and implements rules to check whether or not auditing and threat detection have been enabled for all SQL databases, as described in CIS 4.2.1 and 4.2.2.


j4v:
Great job 😄 

My only comment would be to be careful with service name case (`Sql` and `MONITOR`):

sc_2019-02-05_10h32m24s

sc_2019-02-05_10h32m22s

I also believe the service is named `SQL Database` (https://azure.microsoft.com/en-us/services/sql-database/)?

misg:
I replaced `sql` by `sqldatabase` everywhere and fixed the name formatting issues, @j4v!

ScoutSuite/utils.py
```diff
@@ -30,7 +30,9 @@
     'kubernetesengine': 'Kubernetes Engine',
     'cloudresourcemanager': 'Cloud Resource Manager',
 
+    'monitor': 'Monitor',
```
j4v: 👍

ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
```diff
@@ -1061,7 +1061,10 @@ var make_title = function (title) {
         return 'Cloud Resource Manager';
     } else if (title == 'storageaccounts') {
         return 'Storage Accounts';
-
+    } else if (title == 'monitor') {
```
j4v: 👍

Issue #127, opened 2019-01-30
Refactor the CLI


zer0x64:
Refactor the CLI interface using a subparser for each provider.

zer0x64:
Also, I will try to remove as much of Opinel as possible.

PR #151, opened 2019-02-06
Refactor/127 cli refactoring

zer0x64:
As per #127 

If I did my job right, you should be able to use your old commands by simply removing the --provider switch. This is all documented in the README.md. So, to run an AWS scan with a profile, the command would be:
./Scout.py aws --profile PROFILE_NAME

This code is really error-prone and not tested that much. Especially, I haven't tested GCP at all and with a bunch of AWS switches. It would be useful to have someone on each to test out a bit seeing if something broke.


j4v:
> it seems like the provider used must now be the first argument (which wasn't the case when --provider was used)

@zer0x64 I'm guessing this is because you're now validating the parameters per-provider, instead of the soup of params we had before? I mean it would make sense to have the `--provider` as the first parameter as that helps validate what's passed to the CLI.

> ...speaking of which, should I clone Scout2's wiki? 🤔

Yes I think it would be a good thing to start. Maybe leave of the bits that currently don't apply/work (looking at you ListAll/RulesGenerator!!!).

zer0x64:
> > it seems like the provider used must now be the first argument (which wasn't the case when --provider was used)
> 
> @zer0x64 I'm guessing this is because you're now validating the parameters per-provider, instead of the soup of params we had before? I mean it would make sense to have the `--provider` as the first parameter as that helps validate what's passed to the CLI.

They actually now uses completely independant subparsers! It's works a little bit like git now: You cannot use git -am "commit message" commit, as the -am switch relates to the "commit" parser. The same thing happens here

README.md
```diff
@@ -165,36 +165,45 @@ The following command will provide the list of available command line options:
 
     $ python Scout.py --help
 
+You can also use this for help on a specific provider:
```
Aboisier:
```suggestion
You can also use this to get help on a specific provider:
```

README.md
```diff
@@ -206,21 +215,32 @@ To scan a GCP ...
...
 
-    $ python Scout.py --provider azure --azure-msi
+    $ python Scout.py azure --msi
     
 When using Scout with a Service Principal, you may run Scout using the following command:
 
-    $ python Scout.py --provider azure --azure-service-principal
+    $ python Scout.py azure --service-principal
+
+You can also pass the credentials you want directly with command line arguments. The remaining one will be asked
```
Aboisier:
```suggestion
You can also pass the credentials you want directly with command line arguments. The remaining ones will be asked
```

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
-        azure_auth_params.add_argument('--azure-password',
+                                       dest='username',
+                                       help='Username of the Scout Suite account')
+        azure_auth_params.add_argument('--password',
                                        action='store',
                                        default=None,
-                                       dest='azure_password',
-                                       help='Password of the Azure account')
+                                       dest='password',
+                                       help='Password of the Scout Suite account')
```
Aboisier: Shouldn't this be Azure Account?

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
+                                      action='store_true',
+                                      help='Run Scout Suite with user credentials')
+        azure_auth_params.add_argument('--username',
                                        action='store',
                                        default=None,
-                                       dest='azure_username',
-                                       help='Username of the Azure account')
-        azure_auth_params.add_argument('--azure-password',
+                                       dest='username',
+                                       help='Username of the Scout Suite account')
```
Aboisier: Shouldn't this be Azure Account?

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
+                                      action='store',
+                                      type=argparse.FileType('r'),
+                                      dest='file_auth',
+                                      metavar="FILE",
+                                      help='Run Scout Suite with the specified credential file')
 
-        azure_auth_params.add_argument('--azure-username',
+        azure_auth_modes.add_argument('--user-account',
+                                      action='store_true',
+                                      help='Run Scout Suite with user credentials')
```
Aboisier: Shouldn't this be Azure Account?

j4v:
# General

I've tried this for AWS, GCP and Azure with basic parameters. I really like this PR, makes the CLI much more clean. A couple comments below.

## parameters vs values

Would it be possible to have `--provider` (or `-p`?) before `aws,gcp,azure`, and then have to specify the provider as a value (as currently implement in `master`)?

```
(venv) ~/Git/ScoutSuite (refactor/127-cli-refactoring ✔) ᐅ python Scout.py         
usage: Scout.py [-h] {ruleset,listall,aws,gcp,azure} ...
Scout.py: error: You need to input a module
```

I'd hate to break backwards compatibility over a detail.

Generaly parameters are preceded by `-`/`--`, and values aren't, so it would be `--ruleset`, `--listall` and `--provider` followed by the other parameters/values.

## --help conformity

I really like the new `--help` command but I'm seeing a few inconsistencies:

For azure it's great:

```
(venv) ~/Git/ScoutSuite (refactor/127-cli-refactoring ✘)✹ ᐅ python Scout.py azure --help
[...]
Authentication modes:
  --cli                 Run Scout Suite using configured azure-cli credentials
  --msi                 Run Scout Suite with Managed Service Identity
  --service-principal   Run Scout Suite with an Azure Service Principal
  --file-auth FILE      Run Scout Suite with the specified credential file
  --user-account        Run Scout Suite with user credentials

Authentication parameters:
  --tenant TENANT_ID    Tenant ID of the service principal
  --subscription SUBSCRIPTION_ID
                        Subscription ID of the service principal
  --client-id CLIENT_ID
                        Client ID of the service principal
  --client-secret CLIENT_SECRET
                        Client of the service principal
  --username USERNAME   Username of the Scout Suite account
  --password PASSWORD   Password of the Scout Suite account
```

For GCP shouldn't this be split between `Authentication modes`, `Authentication parameters` (for the SA key file if we keep that parameter) and maybe `Additional arguments` (for project/folder/org filtering):

```
(venv) ~/Git/ScoutSuite (refactor/127-cli-refactoring ✘)✹ ᐅ python Scout.py gcp --help  
[...]
GCP parameters:
  --user-account        Run Scout Suite with a Google Account
  --service-account SERVICE_ACCOUNT
                        Run Scout Suite with a Google Service Account with the
                        specified Google Service Account Application
                        Credentials file
  --project-id PROJECT_ID
                        ID of the GCP Project to analyze
  --folder-id FOLDER_ID
                        ID of the GCP Folder to analyze
  --organization-id ORGANIZATION_ID
                        ID of the GCP Organization to analyze
```

Same for AWS, I'd put these under `Authentication parameters`:

```
  --profile PROFILE [PROFILE ...]
                        Name of the profile. Defaults to ['default'].
  --mfa-serial MFA_SERIAL
                        ARN of the user's MFA device
  --mfa-code MFA_CODE   Six-digit code displayed on the MFA device.
  --csv-credentials CSV_CREDENTIALS
                        Path to a CSV file containing the access key ID and
                        secret key
```

And these under `Additional arguments`:

```
  --regions REGIONS [REGIONS ...]
                        Name of regions to run the tool in, defaults to all
  --vpc VPC [VPC ...]   Name of VPC to run the tool in, defaults to all
  --ip-ranges IP_RANGES [IP_RANGES ...]
                        Config file(s) that contain your known IP ranges
  --ip-ranges-name-key IP_RANGES_NAME_KEY
                        Name of the key containing the display name of a known
                        CIDR
```

# GCP

For service accounts, the parameter is now `--service-account SERVICE_ACCOUNT`. Could you either restore the `--key-file` parameter or change `SERVICE_ACCOUNT` to `KEY_FILE` so it's clear what's required (and update README accordingly)? 

Trying to run for a user account seems to fail (I assume something isn't being passed down the code):

```
(venv) ~/Git/ScoutSuite (refactor/127-cli-refactoring ✘)✹ ᐅ python Scout.py gcp --user-account
Failed to authenticate to GCP - no supported account type
```

README.md
```diff
@@ -165,36 +165,45 @@ The following command will provide the list of available command line options:
...
+    $ python Scout.py aws --help
+
 For further details, checkout our Wiki pages at <https://github.com/nccgroup/ScoutSuite/wiki>.
 
 After performing a number of API calls, Scout will create a local HTML report and open it in the default browser.
 
+Also note that the command line will try to infer the argument name if possible when receiving partial switch. For
+example, this will work and use the selected profile:
+
+    $python Scout.py aws --pro PROFILE
```
j4v: `--profile`

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
+                            help='Do not automatically open the report in the browser.')
+        parser.add_argument('--thread-config',
+                            dest='thread_config',
+                            type=int,
+                            default=4,
+                            help='Level of multi-threading wanted [1-5]; defaults to 4.')
+        parser.add_argument('--report-dir',
+                            dest='report_dir',
+                            default=DEFAULT_REPORT_DIR,
+                            help='Path of the ScoutSuite report.')
```
Aboisier: Some places say `ScoutSuite`, and some other places say `Scout Suite`. We should choose one and be consistent. The Readme uses `Scout Suite`

j4v:
There was a bug in `develop` for GCP service account, fixes has just been merged so you might want to merge develop into this first.

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
+        parser.add_argument('--ip-ranges',
+                            dest='ip_ranges',
+                            default=[],
+                            nargs='+',
+                            help='Config file(s) that contain your known IP ranges')
+        parser.add_argument('--timestamp',
```
Aboisier: This argument is also in the common parser, do we need it here as well?

README.md
zer0x64: We can remove that part, but that was exactly what I was trying to show: The default behavior of argparse is to infer the switch if it cannot find it. So in that case, --pro, --prof, --profi, and so on would all work in place of --profile. But if this is confusing, we can remove it and let people figure it out by themselves!

j4v: Damn, didn't know that! And hadn't read the paragraph above 😅 . So yeah, keep it!

ScoutSuite/__main__.py
```diff
@@ -41,85 +40,79 @@ def main(passed_args=None):
...
+        if args.get('project_id'):
+            report_file_name = 'gcp-%s' % args.get('project_id')
+        elif args.get('organization_id'):
+            report_file_name = 'gcp-%s' % args.get('organization_id')
+        elif args.get('folder_id'):
+            report_file_name = 'gcp-%s' % args.get('folder_id')
         else:
             report_file_name = 'gcp'
     if cloud_provider.provider_code == 'azure':
         report_file_name = 'azure'
```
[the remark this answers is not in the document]
j4v: Yeah that would be a good idea and relates to https://github.com/nccgroup/ScoutSuite/issues/154.

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
+
+    def _init_rules_generator_parser(self):
+        parser = self.subparsers.add_parser("ruleset",
+                                            parents=[self.base_args_parser],
+                                            help="Run the ScoutSuite rules generator")
+        parser.add_argument('--ruleset-name',
+                            dest='ruleset_name',
+                            default=None,
+                            required=True,
+                            help='Name of the ruleset to be generated.')
```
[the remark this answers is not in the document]
j4v: I'd name it something other then `default.json` though, as that's what Scout already uses.

ScoutSuite/cli_parser.py
j4v: Yeah I use `Scout Suite` and `Scout` for short.

zer0x64:
> Would it be possible to have `--provider` (or `-p`?) before `aws,gcp,azure`, and then have to specify the provider as a value (as currently implement in `master`)?

The way the subparsers works(so we can only show the useful help options), we cannot strictly do this. However, a simple fix would be to add a dummy --provider argument before that's useless, but wouldn't break compatibility. However, I don't see how I could do it using argparse so it can be put anywhere(it would still have to be the first parameters.

> Generaly parameters are preceded by `-`/`--`, and values aren't, so it would be `--ruleset`, `--listall` and `--provider` followed by the other parameters/values.

I "could" do the --ruleset and --listall part in a working way, but I'm not sure if we still keep them. Also, t would break compatibility for python 2.6-3.6 as they added the "required" parameter on subparsers in 3.7 and before that it was simply always required.

> For GCP shouldn't this be split between `Authentication modes`, `Authentication parameters` (for the SA key file if we keep that parameter) and maybe `Additional arguments` (for project/folder/org filtering):
> [...]
> Same for AWS, I'd put these under `Authentication parameters`:
> [...]

Will do!

> For service accounts, the parameter is now `--service-account SERVICE_ACCOUNT`. Could you either restore the `--key-file` parameter or change `SERVICE_ACCOUNT` to `KEY_FILE` so it's clear what's required (and update README accordingly)?

The "KEY_FILE" is a really easy fix. If I do that, will I have to restore the --key-file argument? Just thought it would be cleaner to only need one switch in that case.

> Trying to run for a user account seems to fail (I assume something isn't being passed down the code):
> 
> ```
> (venv) ~/Git/ScoutSuite (refactor/127-cli-refactoring ✘)✹ ᐅ python Scout.py gcp --user-account
> Failed to authenticate to GCP - no supported account type
> ```

Is this the fix that has been pushed to develop?

Also, for single letters(like -p for --provider), I can make a bunch of them, as long as they don't conflict(if, let say, I have a -u for --username in azure, it won't conflict with another -u aws-specifc switch). Do you want me to implement a few of them?

ScoutSuite/__main__.py
```diff
@@ -41,85 +40,79 @@ def main(passed_args=None):
     # if not check_requirements(requirements_file_path):
     #     return 42
```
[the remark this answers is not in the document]
zer0x64: lolnope.

ScoutSuite/cli_parser.py
zer0x64: I tried to keep the name consistant with GCP --user-account. Also, since the swtiches are independant, we don't need to precise which provider as they are already provider-specific

ScoutSuite/__main__.py
```diff
@@ -41,85 +40,79 @@ def main(passed_args=None):
...
-                                  profile=args.profile[0],
-                                  project_id=args.project_id,
-                                  folder_id=args.folder_id,
-                                  organization_id=args.organization_id,
-                                  report_dir=args.report_dir,
-                                  timestamp=args.timestamp,
-                                  services=args.services,
-                                  skipped_services=args.skipped_services,
-                                  thread_config=args.thread_config)
+    # TODO fix this mess
```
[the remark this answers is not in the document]
zer0x64: Done!

ScoutSuite/__main__.py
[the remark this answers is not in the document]
zer0x64: Done!

ScoutSuite/cli_parser.py
```diff
@@ -1,264 +1,339 @@
...
+                            help='Ruleset to be used as the baseline.')
+        parser.add_argument('--rules-dir',
+                            dest='rules_dir',
+                            default=[],
+                            nargs='+',
+                            help='Path to directories where custom rules are defined.')
+        parser.add_argument('--generator-dir',
+                            dest='generator_dir',
+                            default=DEFAULT_REPORT_DIR,
+                            help='Path of the Scout2 rules generator.')
```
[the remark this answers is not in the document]
zer0x64: Seems like we use the name Scout now, will implement this on my side

ScoutSuite/cli_parser.py
zer0x64: oups

ScoutSuite/cli_parser.py
zer0x64: Yep, as Listall doesn't use the common parser. I believed it was cleaner to duplicate this one than to create another parser for a single argument

ScoutSuite/cli_parser.py
zer0x64: Done!

ScoutSuite/cli_parser.py
Aboisier: My bad, I copy pasted this comment one too many times haha

ScoutSuite/cli_parser.py
Aboisier: Sounds good!

ScoutSuite/cli_parser.py
zer0x64: Replaced everything with Scout

ScoutSuite/cli_parser.py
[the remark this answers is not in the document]
zer0x64: Not really sure I understand here...

j4v:
> The way the subparsers works(so we can only show the useful help options), we cannot strictly do this.

As we've decided to deprecate listall/rulesgenerator at least for now I think that simplifies the issue. I'm ok with removing the `--provider` parameter and having to provide the cloud provider name as a first argument.

> The "KEY_FILE" is a really easy fix. If I do that, will I have to restore the --key-file argument?

No, you won't have to.

> Just thought it would be cleaner to only need one switch in that case.

I agree, as you will always have to specify the path of the key file when using an SA. So yeah just make it clear that what has to be provided is the path of the key file 😃 

> Is this the fix that has been pushed to develop?

No, I think something might be wrong with how the parameters are passed when using the `--user-account` parameter.

> for single letters

I think it can be useful where it makes sense. The great think with how you've implemented this is that indeed there's less possibility for conflict 👍

zer0x64:
Everything fixed, some arguments also have single-letter switches(I did what I could without causing conflict) and to --provider switch still works withough polluting the help message. Are we ready to merge?

j4v:
> Are we ready to merge?

Please do, I'll try to test as many use cases as I can this week :smiley:

Issue #120, opened 2019-01-25
GCP Multi-Project Scanning with single service account


rtomlinson-latacora:
I would like ScoutSuite to be able to scan multiple projects with a single service account and compile a single report. Much like it does with a user account. Happy to work on this issue. Thank you.

https://cloud.google.com/sdk/gcloud/reference/projects/list
https://stackoverflow.com/questions/35479025/cross-project-management-using-service-account

j4v:
Further discussions in https://github.com/nccgroup/ScoutSuite/issues/80.

@rtomlinson-latacora if you'd be interested in working on this feature, it should involve some [relatively simple] changes to https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/gcp/provider.py. The `_get_projects` currently implements this for user accounts.

rtomlinson-latacora:
Sounds great and thank you. I'll take a look at the code and get a PR to you.

rtomlinson-latacora:
Hmm, so it should be pretty straightforward. The only thing that's causing service accounts to only scan the project in which the service account itself was created is line 157 on https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/gcp/provider.py. It filters based on `parent_id`. Maybe we can default to having a flag for the service account and have all the projects pulled? I'm not too sure how nccgroup likes to do these things so open to suggestions.

rtomlinson-latacora:
I'm assuming that `parent_type` will only be set to `project` if the `--project-id` arg is set on the cli

j4v:
> I'm assuming that `parent_type` will only be set to `project` if the `--project-id` arg is set on the cli

Not exactly, if you run Scout with `--service-account` and provide credentials through the `--key-file` argument, the credentials JSON file contains the `project_id` parameter which is used here.
If you provide the additional `--project-id` argument, that allows setting another project to scan for the SA.

The issue here is that you need a way to determine all the projects to which a SA has access. 
For a user that's pretty easy as if it has IAM privileges at the Organization level, the `_get_projects` method will be able to run `resource_manager_client_v1.projects()` and get a list of all the projects in that Org.

So in order to have similar functionality for a SA, we'd have to implement something that allows enumerating the projects to which the SA has access.

rtomlinson-latacora:
I was able to get scanning all projects that a SA has access to working with my PR the same way that a user account with organization level access does.
[...]

Issue #31, opened 2018-11-27
FN for "No Export Sinks" finding with organizations


j4v:
The `No Export Sinks` finding simply checks that there is a sink configured. When running Scout for an Organization or against multiple projects (i.e. with a Service Account that has access to multiple Projects), if there is 1 sink in the whole Organization/Project list, then the finding will not be flagged.

The finding should check that there is no Project that doesn't export logs to a sink.

j4v:
In order to do this, https://github.com/nccgroup/ScoutSuite/issues/80 will have to be implemented. Currently all the resources are "flat" so it's not possible to write per-project rules.

j4v:
Another issue here is that it's currently not possible to get the `project` attribute for a sink (for example in `parse_sinks` from `stackdriverlogging.py`). 

The reason for this is that in `/ScoutSuite/providers/gcp/configs/base.py` at L178, when the `_client` attribute is set to `None`, this messes with the object's properties and sets them to `NoneType`. 
The Sink properties currently don't have setters (see https://github.com/googleapis/google-cloud-python/blob/master/logging/google/cloud/logging/sink.py) so not sure how to solve this.

Remi05:
This should be fixed with the new architecture. I do not currently have organisation-level to test it and confirm though. @j4v could you check if it is indeed fixed on develop when you have time please?

j4v:
@Remi05 I'll test it out tomorrow but I do believe you have org-level access.

j4v:
Implemented in `develop`.

PR #152, opened 2019-02-06
Feature/120 gcp sa multi project scanning

j4v:
This PR implements fixes for:
- https://github.com/nccgroup/ScoutSuite/issues/31
- https://github.com/nccgroup/ScoutSuite/issues/120


j4v:
@rtomlinson-latacora would greatly appreciate if you can take a look.

rtomlinson-latacora:
@j4v Looks good. I like the changes to the logic in `authenticate`. Any reason that you changed lines 93-94 in `ScoutSuite/providers/gcp/provider.py` to just `gcp` instead of keeping the service account name? I'm not sure how relevant it is but those values show up in the generated HTML report.

j4v:
> Any reason that you changed lines

Didn't smart merge. I've added a commit so that it includes the SA email if there is one. Ideally for a user I'd put his name/email, but I think we'd have to make an extra API query for that - keeping `GCP` for that case for now.

rtomlinson-latacora:
@j4v I just tested this code and service accounts that have an inferred project will never be able to reach the `else` clause to allow for scanning of all projects. I'm of the option that service accounts should scan all projects by default and only a specific project if defined on the CLI, regardless whether or not there is a default project in the credentials file. Let me know what you think. Thanks

j4v:
> service accounts that have an inferred project will never be able to reach the `else`

That's correct: in order to have the SA scan all the accounts to which it has access to, you have to manually remove the default project from the credentials file.

I think that by default, Scout should only scan a user's/SA's default Project. Mainly because in an engagement, you want to make sure you have as little impact as possible. Scanning a client's whole Organization inadvertently might be bad for business!

One thing though is that currently it isn't made very clear what's being scanned, so here's what I propose:
- Adding a bit of output to tell the user what's being scanned
- Add a CLI parameter `--al` to explicitly tell Scout to scan everything the user/SA has access to.

rtomlinson-latacora:
Adding the CLI parameter now.
If we don't want the default behavior to scan all projects, we should probably change the logic in the `ScoutSuite/providers/gcp/provider.py`'s function `authenticate` since the `else` clause essentially defaults to scanning all projects. I'll create a branch and PR again.

j4v:
If you're going to implement this please take https://github.com/nccgroup/ScoutSuite/pull/151 into account, @zer0x64 has be refactoring the CLI.

> If we don't want the default behavior to scan all projects, we should probably change the logic in the `ScoutSuite/providers/gcp/provider.py`'s function `authenticate` since the `else` clause essentially defaults to scanning all projects.

I agree. So maybe by default if we can't determine the default project we can just exit with a message advising the user to either specify a project or run with `--all`?

rtomlinson-latacora:
Looking at #151. Shouldn't be a problem. Are you using a logging library to print messages to stdout?

j4v:
> Are you using a logging library to print messages to stdout?

```
from opinel.utils.console import printInfo
printInfo('string')
```

should do the job

rtomlinson-latacora:
Unless how much information to put on the CLI. We could just show the number of projects or we can display the names of each one. I'm open to either

j4v:
Maybe the number would be enough, otherwise it could look bad for a big org

rtomlinson-latacora:
Just made a PR

PR #153, opened 2019-02-07
feature/#26-new-finding-azure-sql-no-tde-nor-ad-admin-configured

misg:
This PR implements two rules related to SQL databases/servers:
- ensuring that Azure Active Directory Admin is configured for every SQL server (CIS 4.1.8)
- ensuring that data encryption is enabled for all databases (CIS 4.2.6)

### DIFF
--- a/ScoutSuite/providers/azure/rules/rulesets/default.json
+++ b/ScoutSuite/providers/azure/rules/rulesets/default.json
@@ -34,7 +34,7 @@
     "sqldatabase-databases-no-transparent-data-encryption.json": [
       {
         "enabled": true,
-        "level": "danger"
+        "level": "warning"
       }
     ],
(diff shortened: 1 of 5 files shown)


ScoutSuite/providers/azure/rules/rulesets/default.json
```diff
@@ -30,6 +30,18 @@
         "enabled": true,
         "level": "warning"
       }
+    ],
+    "sqldatabase-databases-no-transparent-data-encryption.json": [
+      {
+        "enabled": true,
+        "level": "danger"
```
j4v: I'd lower this to `warning`. While we always recommend encryption on disk, in cloud environment the likeliness of someone accessing the database service directly is low.

PR #155, opened 2019-02-07
Feature/120 gcp sa multi project scanning

Aboisier:
I suggest we push this branch to master as a hotfix since it fixes features broken by the previous release.


Aboisier:
Seems like the bug I was referring to was merged after the release, and is therefore not in master.

PR #159, opened 2019-02-07
Fixed the missing highlighting

Aboisier:
It seems like these were lost in a previous commit, which broke the highlighting.


Aboisier:
> The number of white spaces after a "{" varies, which annoys me, but other than that it's good.

Definitely. I did in the browser, where the automatic formatting could not save me 😢 . I fixed it 👍

Issue #136, opened 2019-02-01
Various UI/UX bugs and enhancements Part 1


j4v:
Font color in the nav bar is a bit off:

<kbd><img src="https://user-images.githubusercontent.com/4206926/52112211-255caa80-2606-11e9-925d-1190e9175742.jpg" /></kbd>

<kbd><img src="https://user-images.githubusercontent.com/4206926/52112212-255caa80-2606-11e9-965c-d2fc4bd45465.jpg" /></kbd>

The badges are a bit off for a number of elements with a "0" count. Ideally a gray badge should be used:

<kbd><img src="https://user-images.githubusercontent.com/4206926/52112213-255caa80-2606-11e9-96a3-bad32b354516.jpg" /></kbd>

<kbd><img src="https://user-images.githubusercontent.com/4206926/52112215-25f54100-2606-11e9-8f06-d29b21fdf989.jpg" /></kbd>

<s>The id_suffix highlighting is broken: </s> (fixed in #159 )

<kbd><img src="https://user-images.githubusercontent.com/4206926/52112217-25f54100-2606-11e9-9862-42f2bbb84d98.jpg" /></kbd>

The loading message is broken:

<kbd><img src="https://user-images.githubusercontent.com/4206926/52230453-b4dcb480-28b7-11e9-81c6-661c3b54ac53.jpg" /></kbd>

j4v:
> The loading message is broken:

Most of the time this will only flash as loading is very quick, but for a large report it's a bit ugly to see.

Aboisier:
Closing, implemented in #167 .

PR #167, opened 2019-02-10
Bugfix/136 various ui bugs part 1

Aboisier:
This PR fixes a couple of UI bugs and inconsistencies described in #136 .

### Inconsistent badges

I created a partial to make the badges uniform. The partial takes a count as input. It will be disabled and greyed out if the count is zero or undefined. It also takes a target as input, in case the badge needs to be linked to an accordion (a lot of badges are used to expand/collapse accordions).
image

### Dark theme menu item color
Self-explanatory.

image

### Please wait modal
Also self-explanatory 

image

### Theme flickering
When loading a huge run_results file, the theme would flicker, so I fixed that as well.


ScoutSuite/output/data/html/partials/metadata.html
```diff
@@ -84,7 +84,7 @@
                   <li><a class="dropdown-item" href="javascript:showLastRunDetails()">Last run details</a></li>
                   <li><a class="dropdown-item" href="javascript:showAbout()">About Scout Suite</a></li>
                   <li><a class="dropdown-item" href="javascript:download_exceptions()">Export exceptions</a><a id="downloadAnchorElem" style="display:none"></a></li>
-                  <li><a class="dropdown-item">Dark theme
+                  <li><a class="dropdown-item" href="#">Dark theme
```
vifor2: Not quite sure about the href, since if the user missclicks while changing the theme he's sent to the Home page, but on the other hand it makes it homogeneous to the other drop-down items :man-shrugging:

ScoutSuite/output/data/html/report.html
```diff
@@ -90,16 +89,18 @@ <h2 id="section_title-h2"></h2>
...
-            <!-- Please wait -->
-            <div class="row" id="please_wait.details">
-                <div class="list-group">
-                    <div class="row list-group-item active">
-                        Please wait
-                    </div>
-                    <div class="row list-group-item">
-                        <p>&nbsp;</p>
-                        <p>Loading data, this may take a while...</p>
-                        <p>&nbsp;</p>
```
vifor2: Good job on removing the &nbsp(s), hate seeing them 👍

j4v:
When middle-clicking (open in new tab) the "Scout Suite" text at top right (I don't think this is a new issue though):

sc_2019-02-11_12h57m28s

When opening a resource/dashboard in a new tab, the loading window is shown (although I'd assume no loading is actually happening):

sc_2019-02-11_10h07m48s

ScoutSuite/output/data/html/partials/metadata.html
Aboisier: You're right, I didn't realize it would do that! I made a fix which keeps it homogeneous but doesn't redirect. It's a bit hacky but at least I don't have to add a css class specifically for this button.

Aboisier:
> When middle-clicking (open in new tab) the "Scout Suite" text at top right (I don't think this is a new issue though):
> 
> sc_2019-02-11_12h57m28s
> 
> When opening a resource/dashboard in a new tab, the loading window is shown (although I'd assume no loading is actually happening):
> 
> sc_2019-02-11_10h07m48s

The run results are loaded every time. If you put a breakpoint in load_metadata, it'll break every time.

j4v:
> The run results are loaded every time. If you put a breakpoint in load_metadata, it'll break every time.

Yeah, you're right. Makes sense, if you open it in another window it's a separate browser context, so the JS variable has to be reloaded. Sorry 😅 .

PR #170, opened 2019-02-11
Fetching termination-protection.

vifor2:
Added 2 rules to CloudFormation service on AWS for issue : https://github.com/nccgroup/ScoutSuite-Proprietary/issues/3 In this public part all that has been added is that I am fetching the value for EnabledTerminationProtection, which was not initially available.

Please see the proprietary part : https://github.com/nccgroup/ScoutSuite-Proprietary/pull/73

image

### DIFF
--- a/ScoutSuite/output/data/html/partials/aws/services.cloudformation.regions.id.stacks.html
+++ b/ScoutSuite/output/data/html/partials/aws/services.cloudformation.regions.id.stacks.html
@@ -19,6 +19,7 @@
         None
       {{/if}}
     </div>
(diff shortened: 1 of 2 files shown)


vifor2:
> I really hate to be that guy but was the view implemented? I noticed you put id suffixes in the proprietary PR, but I saw the implementation of the view neither in this PR nor in the proprietary one.

No worries, thank you watching out. There's already a view implemented for CloudFormation stacks, the policy is visible for stacks who have a policy, but the 'termination_protection' status is not visible there, I'll add it! 
image

vifor2:
image

PR #132, opened 2019-01-31
Feature/61 ui improvements

Aboisier:
This PR addresses a couple of UI issues (#61). There is still quite a lot do to in my opinion, but this takes care a few irritants.

1. Navbar refactoring (you can now click outside to close the navbar
2. The last run data was moved to a modal accessible near the "about section" in the top right corner
3. Expanding a finding by clicking the `+` transforms the said `+` to a `-`, and vice versa
4. The services issues page needed a bit more guidance, I refactored it
5. Added a bit of spacing in the sidebar
6. Refactored the whole modal system. Clicking outside will now close them. They also have a fancy fade-in animation. There is no code-duplication nightmare anymore.

Sorry about all the file changes. They are mostly one-liners though. It's for a good cause.

demo-ui-improvements

PR #172, opened 2019-02-11
feature/#25-new-finding-azure-appgateway-waf-disabled

misg:
This PR adds support for the Application Gateway service.
Please look at the [proprietary PR](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/74) as well.


zer0x64:
Not sure, but this "might" be CIS2.8, as this is also related to a Web Application Firewall(WAF)

ScoutSuite/output/data/html/partials/azure/services.appgateway.app_gateways.html
```diff
@@ -0,0 +1,25 @@
...
+</script>
+
+<script>
+    Handlebars.registerPartial("services.appgateway.app_gateways", $("#services\\.appgateway\\.app_gateways\\.partial").html());
+</script>
+
+<!-- Single Application Gateway template -->
+<script id="single_appgateway_app_gateways-template" type="text/x-handlebars-template">
+    <div style="text-align: right; padding-right: 10px; text-weight: bold;"><a href="javascript:hidePopup()">X</a></div>
+    {{> services.appgateway.app_gateways}}
```
Aboisier: Please prefer using the modal partial instead of this implementation. The javascript:hidePopup does not exist anymore.  See [this](https://github.com/nccgroup/ScoutSuite/pull/132/files#diff-795b3bcacbbb37ba1939693c79cb20a7) example.

misg: Done, sorry about that. :x

requirements.txt
```diff
@@ -24,5 +24,6 @@ oauth2client>=4.1.3
 azure-mgmt-storage>=3.0.0
 azure-mgmt-monitor
 azure-mgmt-sql
+azure-mgmt-network
```
j4v: Same here, please set minimum version (you can use the current version) for all libraries.

misg:
I made the support for Application Gateway service private in the end, since there's no mention of it in CIS current version.

PR #184, opened 2019-02-15
Feature/10 dynamo db findings

vifor2:
Public part of https://github.com/nccgroup/ScoutSuite-Proprietary/issues/10 : 

- Added dynamodb.py to fetch information relevant to the user's dynamoDB service (database tables).
- Added a view for DynamoDB tables, see image below :
dynamo1

Private part here : https://github.com/nccgroup/ScoutSuite-Proprietary/pull/79


vifor2:
> As all the rules are private can you move this to the proprietary repo?

Hmmm sure, just to be sure it's only the html partial here that you want me to move around right? I am asking since some of the files I've modified do not currently exist on the private repo (e.g. services.py).

vifor2:
> As all the rules are private can you move this to the proprietary repo?

Once I move the html partial and dynamodb.py to the private repo the execution of the public repo will fail because service.py will be missing an import so do I also clone service.py into the private repo and each time I add a change to the public part of service.py I also duplicate that change to the private version?

ScoutSuite/providers/aws/services/dynamodb.py
```diff
@@ -0,0 +1,51 @@
...
+        table_details['name'] = table
+
+        table_details['manual_backups'] = api_client.list_backups(TableName=table)['BackupSummaries']
+
+        continous_backup = api_client.describe_continuous_backups(TableName=table)['ContinuousBackupsDescription']
+        continous_backup = continous_backup['PointInTimeRecoveryDescription']['PointInTimeRecoveryStatus']
+        if continous_backup == 'ENABLED':
+            table_details['automatic_backups_enabled'] = True
+        else:
+            table_details['automatic_backups_enabled'] = False
```
Aboisier: Cette ligne pourrait être simplifiée.

`table_details['automatic_backups_enabled'] = continous_backup == 'ENABLED'`

ScoutSuite/providers/aws/services/dynamodb.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self.tables[len(self.tables)] = table_details
```
Aboisier: On utilise l'id de la ressource habituellement plutôt qu'un index (exemples: [AWS Lambda](https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/aws/services/awslambda.py#L18), [CloudFormation](https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/aws/services/cloudformation.py#L31), etc).

```suggestion
        self.tables[table] = table_details
```

ScoutSuite/output/data/html/partials/aws/services.dynamodb.regions.id.tables.html
```diff
@@ -0,0 +1,29 @@
...
+  {{#unless scout2_link}}
+    <div id="resource-name" class="list-group-item active">
+      <h4 class="list-group-item-heading">{{TableName}}</h4>
+    </div>
+    <div class="list-group-item">
+    <h4>Information</h4>
+      <ul>
+        <li class="list-group-item-text">Id: {{TableId}}</li>
+        <li class="list-group-item-text">Arn: {{TableArn}}</li>
+        <li class="list-group-item-text">Automatic backups enabled: {{automatic_backups_enabled}}</li>
```
Aboisier: You're missing the id_suffix.

```suggestion
        <li class="list-group-item-text">Automatic backups enabled: <span id="dynamodb.regions.{{region}}.tables.{{@key}}.automatic_backups_enabled">{{automatic_backups_enabled}}</span></li>
```

ScoutSuite/output/data/html/partials/aws/services.dynamodb.regions.id.tables.html
```diff
@@ -0,0 +1,29 @@
+
+<!-- Table partial -->
+<script id="services.dynamodb.regions.id.tables.partial" type="text/x-handlebars-template">
+  {{#unless scout2_link}}
```
Aboisier: What's that for? I've never seen that!

j4v:
> Once I move the html partial and dynamodb.py to the private repo the execution of the public repo will fail because service.py will be missing an import so do I also clone service.py into the private repo and each time I add a change to the public part of service.py I also duplicate that change to the private version?

So for AWS the logic should be that if there are no public rules for a service (which is the case for dynamodb, right?) then everything should be in the private repo. See the GCP GKE service as an example.

As an example, for GCP GKE:
- the service is loaded with a try/catch clause in https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/gcp/configs/services.py, so there won't be any exceptions thrown in the public version
- resources are included in https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/gcp/metadata.json. in the public version as nothing is pulled, resources will be grayed out in the public version

ScoutSuite/providers/aws/services/dynamodb.py
vifor2: Ohhhh right thanks, I had that weird feeling too when writing it that there's a shorter way to do it!

ScoutSuite/output/data/html/partials/aws/services.dynamodb.regions.id.tables.html
vifor2: To be quite frank, I only put it there because I wanted to keep the same format as previous files and that section is present in many AWS partials, e.g. : services.cloudformation.regions.id.stacks.html, services.cloudtrail.regions.id.trails.html, services.cloudwatch.regions.id.alarms.html, services.ec2.regions.id.volumes.html, services.ec2.regions.id.vpcs.id.images.html and so on...

vifor2:
Public and private PRs updated, thanks a lot for the feedback @j4v & @Aboisier

PR #188, opened 2019-02-17
Feature/32 proprietary azure network findings

zer0x64:
Implementation of the "Network" service and CIS findings section 6 except 6.4. 
Also includes network rules processing for use in "External Attack surface" summary, which will be done in another issue as the GCP and Azure provider doesn't really supports summaries yet.


ScoutSuite/providers/azure/rules/findings/network-security-groups-rule-inbound-ssh.json
```diff
@@ -0,0 +1,13 @@
...
+    "rationale": "You should not permit SSH(port 22) inbound access to a network security group (CIS 6.2).",
+    "path": "network.network_security_groups.id.security_rules.id",
+    "display_path": "network.network_security_groups.id",
+    "conditions": [ "and",
+        ["network.network_security_groups.id.security_rules.id.destination_ports", "containAtLeastOneOf", "22"],
+        ["_INCLUDE_(conditions/exposed-to-the-internet.json)", "", ""],
+        ["_INCLUDE_(conditions/allow-tcp.json)", "", ""]
+    ],
+    "id_suffix": "security_groups_rule_inbound_ssh"
+}
```
Aboisier: It seems like those three rules are very similar, and the only difference is the port, the protocol, and the CIS number. I suggest grouping this all in one rule and using arguments (`__ARG_0__`).

ScoutSuite/providers/azure/services/network.py
```diff
@@ -0,0 +1,149 @@
...
+        self.network_security_groups = {}
+        self.network_security_groups_count = 0
+
+        super(NetworkConfig, self).__init__(thread_config)
+
+    def parse_network_watchers(self, network_watcher, params):
+        network_watcher_dict = {}
+        network_watcher_dict['id'] = network_watcher.id
+        network_watcher_dict['name'] = network_watcher.name
+        network_watcher_dict['provisioning_state'] = network_watcher.provisioning_state == "Succeeded"
```
Aboisier: I find this name a bit confusing, I would expect some kind of state, but it contains a boolean. I'd rename to something like `is_provisioning`, or put the state directly (since it has three possible values: creating, resolvingDNS or succeeded).

ScoutSuite/providers/azure/services/network.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @staticmethod
+    def _merge_prefixes_or_ports(port_range, port_ranges):
+        port_ranges = port_ranges if port_ranges else []
+        if port_range:
+            port_ranges.append(port_range)
+        return port_ranges
+
+
+def _format_ports(ports):
```
Aboisier: To be consistent with your class, I'd put this function in the class.

ScoutSuite/providers/azure/services/network.py
```diff
@@ -0,0 +1,149 @@
...
+                    for p in ports:
+                        exposed_ports.add(p)
+                else:
+                    for p in ports:
+                        exposed_ports.discard(p)
+        exposed_ports = list(exposed_ports)
+        exposed_ports.sort()
+        return exposed_ports
+
+    @staticmethod
```
Aboisier: To be consistent with the rest of the codebase, I'd remove the `@staticmethod`. I agree it makes sense to make those methods static, but `@staticmethod` is used literally once in the whole codebase, so it kind of clashes IMHO.

ScoutSuite/providers/azure/rules/findings/network-security-groups-rule-inbound-ssh.json
zer0x64: Not sure if I'm missing something, but I don't see how I can make separate rules using this, it only takes the last defined one

ScoutSuite/providers/azure/rules/findings/network-security-groups-rule-inbound-service.json
```diff
@@ -0,0 +1,14 @@
+{
+    "dashboard_name": "Network",
```
Aboisier: You can add a `arg_names` rule to provide a description for the arguments.

``` suggestion
    "dashboard_name": "Network",
    "arg_names": [ "Protocol (ex: SSH, RDP, etc.)", "Protocol's port", "Associated CIS rule" ],"
```

zer0x64:
> Whoa, that's a lotta lines there pal

Everybody nows that LOC is the best metric ;)

Issue #181, opened 2019-02-13
Remove Opinel Dependency


zer0x64:
Extract conditions.py from the Opinel library.
After this, it will be possible to add conditions to the rule engine.

j4v:
Should be pretty easy, they are all in https://github.com/nccgroup/opinel/blob/master/opinel/utils/conditions.py.

If you want to merge in Opinel progressively, you could just put all the code in an `opinel` folder in Scout, remove the requirement for the actual library, and voila. From there you can start slowly refactoring from within Scout.

PR #190, opened 2019-02-17
Refactoring/181 Pull Opinel in Scout

zer0x64:
First step of the Opinel refactoring is to pull the code into the Scout code(as mentionned in https://github.com/nccgroup/ScoutSuite/issues/181).

This PR aims to pull Opinel code into Scout Suite so we can modify it and start removing unused componnents.


requirements.txt
```diff
@@ -4,7 +4,10 @@ netaddr>=0.7.11
 
 # AWS Provider
 boto3>=1.9.60
-opinel>=3.3.4,<4.0.0
+
+# Opinel
+requests>=2.4.0,<3.0.0
+iampoliciesgonewild>=1.0.6.2
```
j4v: I didn't know we had this dependency. Hasn't been updated since 2016, I wonder if it supports new services? Could be worth checking where it's used and if the dependency can be removed.

PR #195, opened 2019-02-19
Release fixes

Aboisier:
I had to do three hotfixes because the readme was not showing properly on PyPi. The problem is related to [this issue](https://github.com/pypa/warehouse/issues/3664). The post syntax is a bit weird, but it is compliant with [PEP 440](https://www.python.org/dev/peps/pep-0440/#post-releases).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 include LICENSE
 include README.md
 include requirements.txt
+recursive-include opinel/data *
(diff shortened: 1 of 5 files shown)


zer0x64:
Tests fails? Time to fix the tests then!

j4v:
@Aboisier any way to make sure in the `setup.py` that private files aren't added to the package? Would be nice to have a failsafe.

PR #196, opened 2019-02-19
Feature/11 aws config not enabled

vifor2:
PR for https://github.com/nccgroup/ScoutSuite-Proprietary/issues/11
Please also consult the [private PR](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/90) (most stuff is over there)

In this public part I have : 
- Added the api call in metadata to fetch recorder information for the ConfigService 
- Added try/catch to call private components of ConfigService
- Changed the 2 instances where my service name was messed with and replaced the function with a regex


ScoutSuite/providers/aws/configs/regions.py
```diff
@@ -50,9 +50,8 @@ def __init__(self, service_metadata=None, thread_config=4):
 
         self.regions = {}
         self.thread_config = thread_configs[thread_config]
-        self.service = \
-            type(self).__name__.replace('Config', '').lower()  # TODO: use regex with EOS instead of plain replace
-
+        self.service = re.sub(r'Config$', "", type(self).__name__).lower()
```
Aboisier: 🥂

PR #197, opened 2019-02-19
WIP - Refactoring/azure sqldatabase config

Aboisier:
Michael's refactoring.


misg:
This refactoring is related to the Azure SQL Databaser service.
In this case, we have the following resources:
- (sql) servers
- (sql) databases
- blob auditing policies
- threat detection policies
- replication links
- ad administrators
- transparent data encryption configurations

And the Azure API enforces the following hierarchy:
- (sql) servers
  - ad administrators
  - (sql) databases
    - blob auditing policies
    - threat detection policies
    - replication links
    - transparent data encryption configurations

This means that for example, threat detection policies belong to a sql database which itself belongs to a sql server. So you need the sql server name to retrieve the sql database config and you need the sql server name and the sql database name to retrieve the threat detection policies.

So the current refactoring tries to take this hierarchy into account when retrieving Azure SQL Database config. It basically does so with a Composite pattern: `blob auditing policies`, `threat detection policies`, `replication links`,  and `transparent data encryption configs` are leaves of the `databases` composite node which, with the leaf `ad administrators` , are contained in the `servers` composite node. Each composite node will iterate through its children to fetch their configs and then parse them in order to generate the json used by the rule engine to verify rules.

ScoutSuite/providers/azure/resources/sql/databases.py
```diff
@@ -0,0 +1,69 @@
...
+            self.databases[db.name] = {
+                'id': db.name,
+            }
+
+            # put the following code in a fetch_children() parent method (typical method for a composite node)?
+            for (resource_config, resource_parser) in self.children.values():
+                resource = resource_config(self.resource_group_name, self.server_name, db.name)
+                await resource.fetch_all(credentials)
+                for k, v in resource_parser.__func__(resource).items():
+                    self.databases[db.name][k] = v
```
Aboisier: True, I think this piece of code could be moved up to a `ResourceGroupComposite` class. One thing though is it seems like you're putting the children directly in the parent, which is kind of flattening the composite structure.
 
The tree looks like:
```json
"databases": {
    "some_database": {
        "auditing_enabled": true,
        "threat_detection_enabled": true,
        "replication_configured": true,
        "transparent_data_encryption_enabled": true
    },
    "... more of that"
}
```

But from my understanding, the tree should look like that

```
"databases": {
    "some_database": {
        "BlobAuditingPolicies" {
            "auditing_enabled": true
            "some_policy":  { ... },
            "some_other_policy": { ... }
         },
        "ThreatDetectionPolicies": {
            "threat_detection_enabled": true,
            "some_policy":  { ... },
            "some_other_policy": { ... }
        },
        "ReplicationLinks": {
            "replication_configured": true,
            "some_link_or_whatever...?": { "... }
            "another_link_config_iguess": { ... }
        }  
        "TransparentDataEncryptions": {
            "transparent_data_encryption_enabled": true
            "it_doesnt_even_make_sense_here": { ... }
        }
    },
    "... more of that"
}
```
Of course, this doesn't make sense. I think you went one level deeper than what I had understood. In my mind, we were going to split in different `_________Config` class every time you have an id basically. So let's say you have `ec2.regions.id.vpcs.id.instances.id`, then you have `Ec2Config`, `RegionsConfig`, `VpcsConfig`, and `InstancesConfig`. It seems like you created a config class for each attribute.

Aboisier:
Great work! This is really encouraging, I think our concept has potential.

A few comments on your implementation:
- You put the children parsing methods in DatabasesConfig. I wonder if the parsing is really the parent's responsibility. I was expecting calling `fetch_all` on a resource would set the parsed values to the object itself. Then, the parent just need to keep a handle on its children config.

ScoutSuite/providers/azure/resources/sql/databases.py
```diff
@@ -0,0 +1,69 @@
...
+            'transparent_data_encryption_enabled': config.value.status == "Enabled"
+        }
+
+    # register children resources:
+    children = {
+        'Database Auditing Settings': (DatabaseBlobAuditingPoliciesConfig, _parse_auditing_config),
+        'Database Threat Detection Policies': (DatabaseThreatDetectionPoliciesConfig, _parse_threat_detection_config),
+        'Replication Links': (ReplicationLinksConfig, _parse_replication_links_config),
+        'Transparent Data Encryptions': (TransparentDataEncryptionsConfig, _parse_transparent_data_encryption_config)
+    }
```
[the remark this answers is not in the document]
misg: They're not necessary indeed, it was just for a purpose of documentation. Although, they could be useful for printing/debugging purposes, don't you think?

ScoutSuite/providers/azure/resources/sql/databases.py
misg: You're correct about the trees. They are different because I'm making a difference between the structure of azure configuration (which looks like your 2nd tree, the most nested one) and the json we want to build and pass to the rule engine (which, for obvious reasons, doesn't need to store the entire azure configuration but just some relevant attributes.
That being said, I think flattening the composite structure like I do is related to your question about giving the responsability to a node to parse its children configs.

> In my mind, we were going to split in different `_________Config` class every time you have an id > basically. So let's say you have `ec2.regions.id.vpcs.id.instances.id`, then you have `Ec2Config`, > `RegionsConfig`, `VpcsConfig`, and `InstancesConfig`. It seems like you created a config class for > each attribute.

Yes, I created a config class for each attribute but I did so because those attributes have their own API call and therefore, look like "leaves" (of our composite pattern) to me. But I don't think AWS API is that much granular. Looks like only one API call `describe_instances` is necessary to retrieve all relevant attributes of an instance... So having nothing nested under `InstancesConfig` would indeed makes sense.

misg:
> Great work! This is really encouraging, I think our concept has potential.

Thank you! Yeah, I think so as well!

> * You put the children parsing methods in DatabasesConfig. I wonder if the parsing is really the parent's responsibility. I was expecting calling `fetch_all` on a resource would set the parsed values to the object itself. Then, the parent just need to keep a handle on its children config.

I had the following scenario in mind: you have a resource `A` with two kind of children `B` and `C`. Let's say you want to define a security rule that need information in both children `B` and `C`, how would you combine the relevant information? I thought giving responsability to the parent `A` would easily solve this kind of scenario.
Although, I don't have any concrete example of this problem so I may have "over-engineering" things here... I don't mind moving parsing responsabilities from parents to children, it should make things a bit clearer/easier. 

> I'm also wondering where do you handle the subscription part?

Great catch! I forgot about that and the `resource_group` concept as well. Although, I don't think I can handle several subscriptions since an execution of Scout Suite is only related to one subscription (there's only one subscription id that is being passed to `scout.py` as argument). I'll have to double check!

ScoutSuite/providers/azure/resources/sql/databases.py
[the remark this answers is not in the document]
Aboisier: I see. I'm usually not a fan of leaving unused stuff in case we want to use it. I guess we're refactoring and this is going to be evolving, so I'll let this one slip. 🤸‍♂️

ScoutSuite/providers/azure/resources/sql/databases.py
Aboisier: I see, fair enough. The flattening still bothers me though, it kind of breaks the structure. 

Here are a few solutions

- Use some kind of `SingleConfig` abstract class to make the behavior explicit. 
- Merge the 4 deepest configs into one `DatabaseConfig` class. 

I'll continue thinking about this.

ScoutSuite/providers/azure/resources/sql/databases.py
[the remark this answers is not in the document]
misg: I'm not a fan of leaving unused stuff either. Tbh the primary purpose of those keys was to document my code but since you found that redundant, let's just remove them!

ScoutSuite/providers/azure/resources/sql/databases.py
misg: Just updated the code to make each node take care of its own parsing. The flattening part has changed but it is still not very satisfying to me, I'll keep thinking about it.
I'll had a azure_facade tomorrow, to take care about the API calls!

j4v:
> * I wonder if the parsing is really the parent's responsibility

I'd expect the leaf to do the parsing.

ScoutSuite/providers/azure/resources/sql/databases.py
j4v: > I don't think AWS API is that much granular

Isn't it though? I think that's why there are so many callbacks used in the AWS services.

j4v:
> Scout Suite is only related to one subscription

Would be interesting to see if it's possible to scan a whole "account" (and all it's subscriptions), but I'm not sure Azure's authentication model supports it.

ScoutSuite/providers/azure/configs/services.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, metadata=None, thread_config=4, **kwargs):
 
         self.storageaccounts = StorageAccountsConfig(thread_config=thread_config)
         self.monitor = MonitorConfig(thread_config=thread_config)
-        self.sqldatabase = SQLDatabaseConfig(thread_config=thread_config)
+        self.sqldatabase = SQLDatabaseConfig()
```
j4v: You'll have to make sure the user can still set how many resources the tool consumes.

ScoutSuite/providers/azure/resources/sql/database_blob_auditing_policies.py
```diff
@@ -2,22 +2,26 @@
 
 from azure.mgmt.sql import SqlManagementClient
 
-from ScoutSuite.providers.base.configs.resource_config import ResourceConfig
+from ScoutSuite.providers.base.configs.resources import Resources
```
Aboisier: Shouldn't this be `Resource`, instead of `Resources`?

ScoutSuite/providers/azure/resources/sql/databases.py
Aboisier: Most of the services have "describe_instance", which allows you to get a bunch of configs. From my understanding, with Azure, it seems like you often need to make separate calls, and some calls are dependant on getting info from other calls. It has a deeper hierarchy, basically.

ScoutSuite/providers/azure/resources/sql/database_blob_auditing_policies.py
misg: I replied here: https://github.com/nccgroup/ScoutSuite/pull/201#pullrequestreview-206907757

misg:
> Would be interesting to see if it's possible to scan a whole "account" (and all it's subscriptions), but I'm not sure Azure's authentication model supports it.

Indeed, I'll look into this! (Although I have the same feeling as yours.)

ScoutSuite/providers/base/configs/resources.py
```diff
@@ -8,3 +8,27 @@ class Resources(dict, metaclass=abc.ABCMeta):
     @abc.abstractmethod
     async def fetch_all(self, **kwargs):
         raise NotImplementedError()
+
+
+class SimpleResources(Resources, metaclass=abc.ABCMeta):
+
```
Aboisier: Would you mind adding docstrings to SimpleResource and CompositeResource?

ScoutSuite/providers/azure/resources/resources.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class AzureCompositeResources(CompositeResources, metaclass=abc.ABCMeta):
+
+    # TODO: get rid of the credentials.
+    async def fetch_children(self, parent, **kwargs):
+        for child_class in self.children:
+            child = child_class(**kwargs)
+            await child.fetch_all()
+            parent.update(child)
```
Aboisier: I think this is generic enough to be moved to the CompositeResources class.

ScoutSuite/providers/azure/resources/resources.py
```diff
@@ -0,0 +1,24 @@
...
+from ScoutSuite.providers.base.configs.resources import SimpleResources
+from ScoutSuite.providers.base.configs.resources import CompositeResources
+
+
+class AzureSimpleResources(SimpleResources, metaclass=abc.ABCMeta):
+
+    async def fetch_all(self):
+        raw_resource = await self.get_resources_from_api()
+        name, resource = self.parse_resource(raw_resource)
+        self[name] = resource
```
Aboisier: I'm pretty confused about this method. Doesn't `get_resources_from_api` return an array of raw resources?

ScoutSuite/providers/azure/resources/sql/server_azure_ad_administrators.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, resource_group_name, server_name, facade):
+        self.resource_group_name = resource_group_name
+        self.server_name = server_name
+        self.facade = facade
+
+    # TODO: make it really async.
+    async def fetch_all(self):
+        try:
+            _ = self.facade.server_azure_ad_administrators.get(self.resource_group_name, self.server_name)
```
Aboisier: Do you need to store the result in a variable?

ScoutSuite/providers/azure/resources/resources.py
misg: It returns an array of raw resources in your case, with AWS. ;) Azure mostly returns a single value at the level of `SimpleResources`... That's why I decided to leave the classes `SimpleResources` and `CompositeResources` abstract in `base/configs/resources.py` and define `AzureSimpleResources` and `AzureCompositeResources`.

ScoutSuite/providers/azure/resources/resources.py
misg: I would wait your EC2 refactoring to see if you need exactly the same for AWS or not.

Aboisier: Sounds good!

Aboisier: I'm concerned about thread safety here. Maybe we should put a comment or something so we don't forget.

ScoutSuite/providers/base/configs/resources.py
misg: Done! Feel free to modify them, I did my best to be concise and clear but they may be still confused!

ScoutSuite/providers/azure/resources/sql/server_azure_ad_administrators.py
misg: Haha, nice catch.

ScoutSuite/providers/azure/resources/resources.py
misg: It shouldn't be a concern because of the famous GIL of CPython (that makes Python "so slow" :stuck_out_tongue: ): https://docs.python.org/3/glossary.html#term-global-interpreter-lock.

Aboisier: That's what I thought! 👍

ScoutSuite/providers/azure/resources/resources.py
```diff
@@ -10,8 +10,7 @@ class AzureSimpleResources(SimpleResources, metaclass=abc.ABCMeta):
 
     async def fetch_all(self):
         raw_resource = await self.get_resources_from_api()
-        name, resource = self.parse_resource(raw_resource)
-        self[name] = resource
+        self.parse_resource(raw_resource)
```
Aboisier: I don't want to be this guy, but I think I preferred the previous implementation. The reason is that `parse_resource` really sounds like it's going to return a parsed resource. It doesn't sound like it's going to sneakily add the resource to the dictionary. I think we should either rename the method, split the two responsibilities into two different methods, or fall back to the initial implementation.

One advantage of having a `parse_resource` returning a parsed resource and fetch_all handling the dictionary is that you reduce code duplication (you don't need to do `self.update` in every `parse_resource`).

j4v: > It doesn't sound like it's going to sneakily add

That's one of my major griefs with the legacy code - a lot of methods have obscure names or are leveraged for operations that aren't immediately clear. bytes are cheap so don't be scared to use verbose names, and please add docstrings to clearly state what these classes/methods/functions are for.

misg:
I'm closing this PR since its git history has been corrupted and we're not able to clean this up. Please look at #214 which is the new clean version of the PR.

PR #199, opened 2019-02-20
WIP - Refactoring/aws lambda config

Aboisier:
Here's a work in progress for migrating the Aws Lambda service to the new architecture. I had to implement a `RegionConfig`, which is generic. I think it's not too ugly.


ScoutSuite/providers/aws/configs/regions_config.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from ScoutSuite.providers.base.configs.resource_config import ResourceConfig
+from ScoutSuite.providers.aws.aws_facade import AwsFacade
+
+
+class RegionsConfig(ResourceConfig):
```
misg: I'm not really convinced about the design of `RegionsConfig` and its "child" `LambdaServiceConfig`. It looks like you're not really taking advantages of our composite structure and inheritance principles, by having `RegionsConfig` **and** `LambdaServiceConfig` inherited from `ResourceConfig`, with `RegionsConfig` instantiated in `LambdaServiceConfig` with `LambdasConfig` passed as an argument.

Why not defining `RegionsConfig` as a base class, something like:

```python
class RegionsConfig(ResourceConfig):
    # TODO: add service as arg
    async def fetch_all(self, credentials, regions=None, partition_name='aws', targets=None):
        # TODO: Should be injected
        facade = AwsFacade()
        self['regions'] = await facade.build_region_list(service, regions, partition_name)
        self['regions_count'] = len(self['regions'])
```

and then just defining `LambdasConfig` as a child class of `RegionsConfig` (and removing `LambdaServiceConfig` which doesn't seem very useful, even though probably used in the current codebase), as following:

```python
class LambdasConfig(RegionsConfig):
    async def fetch_all(self, credentials, region, partition_name='aws', targets=None):
        super(LambdasConfig, self).fetch_all('lambda', region, partition_name) # TODO: modify fetch_all to match this signature

        # TODO: Should be injected
        facade = AwsFacade()
        for region in self['regions']:
                functions = {}
                for raw_function in facade.get_lambda_functions(region):
                    name, function = self.parse_function(raw_function)
                    functions[name] = function

        self['functions_count'] = len(functions)        
        self['functions'] = functions

    @staticmethod
    def parse_function(function):
        function['name'] = function.pop('FunctionName')
```
But we would have to deal with two things first, I believe:
- removing `LambdaServiceConfig` safely, ensuring compliance with the interface of the current codebase (mainly with the rule engine, in the end)
- modifiying `fetch_all` signature, to be able to pass a `service` argument. More than that, it seems like you don't need `credentials` and `targets` argument. I would suggest to modify the base class `ResourceConfig` to be more general:
```python
class ResourceConfig(dict):

        raise NotImplementedError()

    async def finalize(self):
        pass

File: ScoutSuite/providers/aws/services/awslambda.py [hunk 2]
```diff
@@ -1,34 +1,30 @@
...
+    async def fetch_all(self, credentials, region, partition_name='aws', targets=None):
+        # TODO: Should be injected
+        facade = AwsFacade()
+        
+        functions = {}
+        for raw_function in facade.get_lambda_functions(region):
+            name, function = self.parse_function(raw_function)
+            functions[name] = function
+
+        self.setdefault('functions_count', len(functions))
```
misg: `self['functions_count'] = len(functions)` would be more "Pythonic", I believe? Since `ResourceConfig` inherits from `dict`.

ScoutSuite/providers/aws/aws_facade.py
```diff
@@ -0,0 +1,34 @@
...
+from botocore.session import Session
+from collections import Counter
+
+class AwsFacade(object):
+    def get_lambda_functions(self, region):
+        aws_lambda = boto3.client('lambda', region_name=region)
+        
+        functions = []
+        
+        marker = None
```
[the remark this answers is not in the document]
Aboisier: Friggin python with its weird scopes. I guess it's not necessary.

ScoutSuite/providers/aws/aws_facade.py
```diff
@@ -0,0 +1,34 @@
...
+        service = 'ec2containerservice' if service == 'ecs' else service
+        available_services = Session().get_available_services()
+
+        if service not in available_services:
+            raise Exception('Service ' + service + ' is not available.')
+
+        regions = Session().get_available_regions(service, partition_name)
+
+        if chosen_regions:
+            return list((Counter(regions) & Counter(chosen_regions)).elements())
```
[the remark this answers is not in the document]
Aboisier: Thanks <3 It comes from opinel. I feel like this line could be simplified to this:

```suggestion
            return list((Set(regions) & Set(chosen_regions))
```

I didn't do it cause I'm not sure.

ScoutSuite/providers/aws/aws_facade.py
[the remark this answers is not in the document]
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/configs/regions_config.py
Aboisier: My idea was really trying to respect the composite structure as much as possible. I like the idea of making RegionConfig a base class. I feel like having one class handle two levels of the tree (in this case _regions_ and _functions_) is somewhat against the purpose of having a composite pattern, because you are hiding multiple layers of the structure in one node. Splitting the service/region-specific configs in one class and the instance-specific configs in another one is a sweet spot.

I'll explore a few different implementations and push the most interesting here for review.

> * modifiying `fetch_all` signature, to be able to pass a `service` argument. More than that, it seems like you don't need `credentials` and `targets` argument. I would suggest to modify the base class `ResourceConfig` to be more general

I agree, the fetch_all signature should be more general.

ScoutSuite/providers/aws/services/awslambda.py
Aboisier: Sotrue.
```suggestion
       self['functions_count'] = len(functions))        
```

ScoutSuite/providers/aws/configs/regions_config.py
misg: > My idea was really trying to respect the composite structure as much as possible. I like the idea of making RegionConfig a base class. I feel like having one class handle two levels of the tree (in this case _regions_ and _functions_) is somewhat against the purpose of having a composite pattern, because you are hiding multiple layers of the structure in one node. Splitting the service/region-specific configs in one class and the instance-specific configs in another one is a sweet spot.
> 
> I'll explore a few different implementations and push the most interesting here for review.

I really like your new implementation! I think I finally got your point about having two separated classes, I was mainly confused by the terminology being used, I believe. Would you mind if I suggest some changes in the name of your classes, to make things clearer?

> I agree, the fetch_all signature should be more general.

Done, I just pushed a more general `fetch_all` method in `refactoring/resource_config`.

ScoutSuite/providers/aws/configs/regions_config.py
```diff
@@ -0,0 +1,19 @@
...
+
+    async def fetch_all(self, chosen_regions=None, partition_name='aws'):
+        # TODO: Should be injected
+        facade = AwsFacade()
+
+        self['regions'] = {}
+        for region in await facade.build_region_list(self._service, chosen_regions, partition_name):
+            self['regions'][region] = {}
+            
+        self.setdefault('regions_count', len(self['regions']))
```
misg: `self['regions_count'] = len(self['regions'])`

ScoutSuite/providers/aws/services/awslambda.py
```diff
@@ -1,34 +1,39 @@
...
-
-class LambdaRegionConfig(RegionConfig):
-
-    def parse_function(self, global_params, region, function):
+from ScoutSuite.providers.base.configs.resource_config import ResourceConfig
+from ScoutSuite.providers.aws.configs.regions_config import RegionsConfig
+from ScoutSuite.providers.aws.aws_facade import AwsFacade
+from opinel.utils.aws import build_region_list
+
+class LambdaServiceConfig(RegionsConfig):
```
misg: Since this classe stores lambdas accross all regions, I would name it `Lambdas`.

ScoutSuite/providers/aws/services/awslambda.py
```diff
@@ -1,34 +1,39 @@
...
+            partition_name=partition_name
+        )
+
+        for region in self['regions']:
+            functions = LambdasConfig()
+            await functions.fetch_all(region=region)
+            self['regions'][region] = functions
+        
+
+class LambdasConfig(ResourceConfig):
```
misg: And since this classe stores lambdas of a specific region, I would call it `RegionalLambdas`.

ScoutSuite/providers/aws/services/awslambda.py
```diff
@@ -1,34 +1,39 @@
...
+        # TODO: Should be injected
+        facade = AwsFacade()
+        
+        functions = {}
+        for raw_function in facade.get_lambda_functions(region):
+            name, function = self.parse_function(raw_function)
+            functions[name] = function
+
+        self['functions_count'] = len(functions))        
+        self.setdefault('functions', functions)
```
misg: `self['functions'] = functions`

ScoutSuite/providers/aws/services/awslambda.py
```diff
@@ -1,34 +1,39 @@
...
+from ScoutSuite.providers.base.configs.resource_config import ResourceConfig
+from ScoutSuite.providers.aws.configs.regions_config import RegionsConfig
+from ScoutSuite.providers.aws.aws_facade import AwsFacade
+from opinel.utils.aws import build_region_list
+
+class LambdaServiceConfig(RegionsConfig):
+    def __init__(self):
+        super(LambdaServiceConfig, self).__init__('lambda')
+
+    async def fetch_all(self, credentials, regions=None, partition_name='aws', targets=None):
```
misg: I think you could get rid of `credentials` and `targets`.

ScoutSuite/providers/aws/configs/regions_config.py
Aboisier: > I think I finally got your point about having two separated classes [...]

I think too! Great teamwork ✋ 

> Would you mind if I suggest some changes in the name of your classes, to make things clearer?

On the contrary; I'm not satisfied with my classes names. I know we talked about removing the `Config` suffix. This would give us something like: `Resource`,  `Region`, `LambdaService`, `LambdaFunction`. Did you have another idea in mind?

ScoutSuite/providers/aws/services/awslambda.py
Aboisier: So true

Aboisier:
```suggestion
    async def fetch_all(self, regions=None, partition_name='aws'):
```

ScoutSuite/providers/aws/aws_facade.py
```diff
@@ -0,0 +1,34 @@
+import boto3
```
j4v: why not just `facade.py`?

ScoutSuite/providers/aws/aws_facade.py
```diff
@@ -0,0 +1,34 @@
...
+from botocore.session import Session
+from collections import Counter
+
+# TODO: Handle authentication better. I don't even know how it currently works. I think connect_service is called somewhere.
```
j4v: It's handled by the `authenticate` method of each provider (e.g. https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/aws/provider.py#L52). For AWS we're passing the CLI params to Opinel.

ScoutSuite/providers/aws/aws_facade.py
```diff
@@ -0,0 +1,34 @@
...
+class AwsFacade(object):
```
j4v: Can we keep the provider names uppercase for consistency (`AWS`, `GCP`, `Azure`)? Details, details...

ScoutSuite/providers/aws/configs/services.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, metadata=None, thread_config=4, **kwargs):
         self.emr = EMRConfig(metadata['analytics']['emr'], thread_config)
         self.iam = IAMConfig(thread_config)
         self.kms = KMSConfig(metadata['security']['kms'], thread_config)
-        self.awslambda = LambdaConfig(metadata['compute']['awslambda'], thread_config)
```
j4v: Out with the metadata ⛏ ⛏ ⛏

ScoutSuite/providers/aws/aws_facade.py
Aboisier: Of course, I never noticed AWSProvider was uppercase, my bad.

ScoutSuite/providers/aws/aws_facade.py
Aboisier: Yea I guess this would be more consistent with the codebase. It's so annoying though when you use the fuzzy search to open a file and it suggests 4 different files all named `provider.py` 😆

Remi05:
Great work! I feel like this new architecture will really make the code more maintainable and easier to understand going forward! :thumbsup:

ScoutSuite/providers/aws/configs/regions_config.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from ScoutSuite.providers.base.configs.resources import Resources
+from ScoutSuite.providers.aws.facade import AWSFacade
+
+
+class RegionsConfig(Resources):
```
misg: Don't you want to rename this class `Regions`? (Just for being consistent with `Resources`, `Lambdas` and `RegionalLambdas`.)

Aboisier: It was renamed in my EC2 PR <3

Aboisier:
Thanks for reviewing and all! I'm just going to close this though as most of the changes are contained in #202 and it has a whole lot of conflicts with this branch.

Issue #183, opened 2019-02-15
Formalize the providers architecture


Aboisier:
_Once upon a time, there was only AWS. The realm of Scout was dominated by this provider only, and as it grew bigger, jealous grew the other providers. GCP was then given a piece of land. GCP, receiving a lot of attention from the Supreme Beings of Scout, was making Azure jealous. And so Azure was given a piece of land. Even though it seemed like the providers were happily cohabiting, the Supreme Beings knew something was not right._ 

And this is why we need to refactor the providers' architecture. Basically, there is a whole lot of provider-checking conditions, which is a huge code smell. There is already a provider model in place, but a lot of attributes should be bubbled up to the generic provider class, and some others should be pushed down to their respective provider. Basically, we should take advantage of polymorphism.

#### AWS
- [x] ~Lambdas **(#199)**~
- [x] ~CloudFormation **(#216)**~
- [x] ~CloudTrail **(#219)**~
- [x] ~CloudWatch **(#221)**~
- [x] ~DirectConnect **(#222)**~
- [x] ~EC2 **(#202)**~
- [x] ~EFS **(#227)**~
- [x] ~ElastiCache **(#245)**~
- [x] ~ELB **(#257)**~
- [x] ~ELBv2 **(#259 )**~
- [x] ~EMR **(#248)**~
- [x] ~IAM **(#264)**~
- [x] ~RDS **(#271 )**~
- [x] ~RedShift **(#261)**~
- [x] ~Route53 **(#267)**~
- [x] ~S3 **(#270)**~
- [x] ~SES **(#269)**~
- [x] ~SNS **(#265)**~
- [x] ~SQS **(#262)**~
- [x] ~VPC **(#273)**~
- [x] ~Config **(#272 and https://github.com/nccgroup/ScoutSuite-Proprietary/pull/115)**~
- [x] ~DynamoDb **(#274 and https://github.com/nccgroup/ScoutSuite-Proprietary/pull/115)**~
- [x] ~KMS (**#280**)~
- [x] ~Clean `metadata.json`~ 

#### Azure
- [x] ~App Gateway **(https://github.com/nccgroup/ScoutSuite-Proprietary/pull/109)**~
- [x] ~App Service **(https://github.com/nccgroup/ScoutSuite-Proprietary/pull/110)**~
- [x] ~Key Vault **(#239)**~
[...]

PR #202, opened 2019-02-22
EC2 migration to the new architecture

Aboisier:
This PR is a work in progress for refactoring the EC2 service, as part of the #183 issue.

I was doing a lot of copy-pasting in between the Lambdas service and this one, and between the facade methods, so I generalized some classes and methods.

EC2 has a pretty complex structure, but it seems like our new architecture holds up.

* EC2
  * regions
    * region_id
      * images *(contains a dictionary of images)*
      * snapshots *(contains a dictionary of snapshots )*
      * volumes *(contains a dictionary of volumes )*
      * vpcs
        * vpc_id
          * instances *(contains a dictionary of instances)*
          * network_interface *(contains a dictionary of nics)*
          * security groups *(contains a dictionary of sgs)*

### `Regions`
Fetches a list of all the regions and builds a `regions` dictionary with <region_id, empty_dictionary> key pairs. It also sets the `regions_count`

### `ScopedResource`
Abstract class which fetches resources in a given scope. The child class needs to define a `get_resources_in_scope(scope)` method. The scope can be anything; a region, some VPC id, etc. It also needs to define a `parse_resource(resource)` method which returns a (resource_id, resource) tuple. The ScopedResource class loops and sets the resources in its own dictionary.


ScoutSuite/providers/aws/configs/regions_config.py
```diff
@@ -15,5 +15,28 @@ def __init__(self, service):
...
+        self.key = key
+
+    async def fetch_all(self, scope):
+        resources = {}
+        for raw_resource in await self.get_resources_in_scope(scope):
+            name, ressource = self.parse_resource(raw_resource)
+            resources[name] = ressource
+
+        self[self.key + '_count'] = len(resources)
+        self[self.key] = resources
```
Aboisier: Not sure this is the best solution. Might also just do `self[name] = resource`, and have the class using a `ScopedResources` handle the count/dic name.

ScoutSuite/providers/aws/provider.py
```diff
@@ -79,12 +79,12 @@ def preprocessing(self, ip_ranges=None, ip_ranges_name_key=None):
...
-        self._add_cidr_display_name(ip_ranges, ip_ranges_name_key)
-        self._merge_route53_and_route53domains()
-        self._match_instances_and_roles()
-        self._match_iam_policies_and_buckets()
+        # self._add_last_snapshot_date_to_ec2_volumes()
+        # self._process_cloudtrail_trails(self.services['cloudtrail'])
+        # self._add_cidr_display_name(ip_ranges, ip_ranges_name_key)
+        # self._merge_route53_and_route53domains()
+        # self._match_instances_and_roles()
+        # self._match_iam_policies_and_buckets()
```
Aboisier: Those currently crash because I'm still missing a lot of stuff in the data structure.

ScoutSuite/providers/aws/services/ec2.py
```diff
@@ -1,330 +1,402 @@
...
+#     Ensure name and ID are set
+
+#     :param vpc_info:
+#     :param vpc_id:
+#     :return:
+#     """
+#     manage_dictionary(vpc_info, vpc_id, {})
+#     vpc_info[vpc_id]['id'] = vpc_id
+#     if not 'name' in vpc_info[vpc_id]:
+#         vpc_info[vpc_id]['name'] = vpc_id
```
Aboisier: Need to split this in multiple classes.

ScoutSuite/output/data/html/partials/aws/services.ec2.regions.id.vpcs.id.instances.html
```diff
@@ -9,7 +9,6 @@ <h4>Information</h4>
                 <li class="list-group-item-text">Region: {{region}}</li>
                 <li class="list-group-item-text">VPC: {{get_value_at 'services.ec2.regions' region 'vpcs' vpc 'name'}} ({{vpc}})</li>
                 <li class="list-group-item-text">ID: {{id}}</li>
-                <li class="list-group-item-text">Reservation ID: {{reservation_id}}</li>
```
Aboisier: I'll need to look into injecting this id to the model.

Aboisier:
I mostly finished migrating EC2 to the new architecture. Something weird is happening though. If you look at the picture below, you can notice the number of resources is different. The previous version showed 17 resources, the new implementation shows 14. The weird thing is they have the same number of individual resources in the navbar, and the sum of the individual resources counts is 14.

As you can see in [this diff](https://www.diffchecker.com/XNXxkSAZ), there are only a few difference between the original result and the refactored result. A few things were removed (aka some multi-threading attributes which were somehow part of the run results), and a few ids were added (because of the way some requests are made now). I also removed "resources_types", I'm not sure it was used.

image

image

ScoutSuite/providers/aws/services/ec2/service.py
```diff
@@ -0,0 +1,35 @@
...
+# TODO: Add docstrings
+
+class EC2(Regions):
+    def __init__(self):
+        super(EC2, self).__init__('ec2')
+
+    async def fetch_all(self, credentials=None, regions=None, partition_name='aws'):
+        await super(EC2, self).fetch_all(chosen_regions=regions, partition_name=partition_name)
+
+        # TODO: Is there a way to generalize this?
```
misg: From what I'm seeing here in your PR, I have the feeling that we share a lot of things in the structure of our providers (aws/azure). So I tried to generalize your `ScopedResources` class in my PR (I renamed your class `SimpleResources`) to extend our composite pattern and use a parent/children structure (see https://github.com/nccgroup/ScoutSuite/pull/197/files).

So if you define a `AwsSimpleResources` and a `AwsCompositeResources` like I did for Azure (https://github.com/nccgroup/ScoutSuite/blob/dcf6e3a0196de64c5aaf4ee9864cb5189c47b81c/ScoutSuite/providers/azure/resources/resources.py), I think you could rewrite this `EC2` class to something like:

```python
class EC2(AwsCompositeResources):

    children = [
        Vpcs,
        Snapshots,
        Volumes,
        AmazonMachineImages
    ]
    def __init__(self):
        super(EC2, self).__init__('ec2')

    async def fetch_all(self, credentials=None, regions=None, partition_name='aws'):
        await super(EC2, self).fetch_all(chosen_regions=regions, partition_name=partition_name)
        for region in self['regions']:
            await self.fetch_children(
                parent=self['regions'][region],
                region=region
            )
             # There could be a way to 'generalize' the following lines too, through a callback system but
            # let's keep it simple for now (and see if such system could be useful, after migrating other AWS services):
            self['regions'][region]['instances_count'] = sum([vpc['instances'].count for vpc in self['regions'][region]['vpcs'].values()])
            self['regions'][region]['security_groups_count'] = sum([vpc['security_groups'].count for vpc in self['regions'][region]['vpcs'].values()])
            self['regions'][region]['network_interfaces_count'] = sum([vpc['network_interfaces'].count for vpc in self['regions'][region]['vpcs'].values()])
```

Remi05:
> ### `ScopedResource`
> Abstract class which fetches resources in a given scope. The child class needs to define a `get_resources_in_scope(scope)` method. The scope can be anything; a region, some VPC id, etc. It also needs to define a `parse_resource(resource)` method which returns a (resource_id, resource) tuple. The ScopedResource class loops and sets the resources in its own dictionary.

What if there are different types of scopes? How would that work?

ScoutSuite/providers/aws/facade/utils.py
```diff
@@ -0,0 +1,15 @@
+class AWSFacadeUtils:
+    @staticmethod
+    def get_all_pages(api_call, parse_response):
+        resources = []
+
+        while True:
+            response = api_call()
+            resources.extend(parse_response(response))
+
+            # TODO: this marker should be passed to te api call. Also, some calls return a NextMarker, some return a NextToken.
```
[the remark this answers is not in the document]
Aboisier: I'll look into this! I doubt it though, my account has a very limited number of resources, I don't think I have multiple pages.

ScoutSuite/providers/aws/configs/regions_config.py
```diff
@@ -14,6 +14,29 @@ def __init__(self, service):
...
+                'name': region
+            }
+
         self['regions_count'] = len(self['regions'])
+
+
+class ScopedResources(Resources):
+    async def fetch_all(self, scope):
+        for raw_resource in await self.get_resources_in_scope(scope):
+            name, ressource = self.parse_resource(raw_resource)
```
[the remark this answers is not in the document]
Aboisier: Omg, classic frenchboi mistake

ScoutSuite/providers/aws/configs/regions_config.py
```diff
@@ -14,6 +14,29 @@ def __init__(self, service):
...
+            self[name] = ressource
+
+        self.count = len(self)
```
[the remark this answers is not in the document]
Aboisier: 🤣 wow that's totally true

j4v:
> you can notice the number of resources is different

Looking from L2192 downwards, you can see that `"snapshots_count": 2` and `"volumes_count": 1` are missing on the right, which would account for the 3 resources missing.

j4v:
Great job to all of you, happy to see you're working closely to implement a multi-provider solution.

Aboisier:
> > ### `ScopedResource`
> > Abstract class which fetches resources in a given scope. The child class needs to define a `get_resources_in_scope(scope)` method. The scope can be anything; a region, some VPC id, etc. It also needs to define a `parse_resource(resource)` method which returns a (resource_id, resource) tuple. The ScopedResource class loops and sets the resources in its own dictionary.
> 
> What if there are different types of scopes? How would that work?

The scope is the info the class needs to fetch the resources. It's called `SimpleResource` now, with @misg  refactoring. It can be a region, a user id, a project id, a vpc id, etc.

ScoutSuite/providers/aws/resources/resources.py
```diff
@@ -0,0 +1,54 @@
...
+            parent[key].update(child)
+            parent[key + '_count'] = len(child)
+
+class Regions(AWSCompositeResources, metaclass=abc.ABCMeta):
+    def __init__(self, service):
+        self.service = service
+        # TODO: Should be injected
+        self.facade = AWSFacade()
+
+    async def fetch_all(self, credentials, chosen_regions=None, partition_name='aws'):
```
misg: I think you wanted your `# TODO: get rid of the credentials.` here.

ScoutSuite/providers/aws/services/ec2/service.py
```diff
@@ -0,0 +1,35 @@
...
+from ScoutSuite.providers.aws.services.ec2.volumes import Volumes
+
+from opinel.utils.aws import get_aws_account_id
+from ScoutSuite.utils import get_keys, ec2_classic
+
+
+# TODO: Add docstrings
+
+class EC2(Regions):
+    children = [
```
misg: yaaaay :fireworks:

ScoutSuite/providers/aws/services/ec2/snapshots.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Snapshots(AWSSimpleResources):
+    async def get_resources_from_api(self):
+        return self.facade.ec2.get_snapshots(self.scope['region'], self.scope['owner_id'])
+
+    def parse_resource(self, raw_snapshot):
+        raw_snapshot['id'] = raw_snapshot.pop('SnapshotId')
+        raw_snapshot['name'] = get_name(raw_snapshot, raw_snapshot, 'id')
+        raw_snapshot['public'] = self._is_public(raw_snapshot)
+        return raw_snapshot['id'], raw_snapshot
```
misg: I'm a bit confused here, do you really need to store `raw_snapshot` instead of only keeping its id, name and public status?

ScoutSuite/providers/aws/services/ec2/volumes.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Volumes(AWSSimpleResources):
+    async def get_resources_from_api(self):
+        return self.facade.ec2.get_volumes(self.scope['region'])
+
+    def parse_resource(self, raw_volume):
+        raw_volume['id'] = raw_volume.pop('VolumeId')
+        raw_volume['name'] = get_name(raw_volume, raw_volume, 'id')
+        return raw_volume['id'], raw_volume
```
misg: Same question as above for `raw_volume`.

ScoutSuite/providers/aws/services/ec2/vpcs.py
```diff
@@ -0,0 +1,32 @@
+from ScoutSuite.providers.aws.resources.resources import AWSCompositeResources, AWSSimpleResources
+from ScoutSuite.providers.aws.facade.facade import AWSFacade
+from ScoutSuite.providers.aws.services.ec2.instances import EC2Instances
+from ScoutSuite.providers.aws.services.ec2.securitygroups import SecurityGroups
+from ScoutSuite.providers.aws.services.ec2.networkinterfaces import NetworkInterfaces
+
+
+class Vpcs(AWSCompositeResources, AWSSimpleResources):
```
misg: Ah, this is weird (but I understand why you're doing this).
Do you think we should merge the abstract classes `Resources` and `SimpleResources` (into `Resources`) to be more coherent? I think it would make sense but will require an other (hopefully final) wave of (small) changes. :face_with_head_bandage: This means `Resources` will define `fetch_all`, `get_resources_from_api` and `parse_resource` and thus `CompositeResources` will inherit `get_resources_from_api` and `parse_resource`. Would that make sense in the case of AWS? (It does with Azure.)

ScoutSuite/providers/aws/services/awslambda/service.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from ScoutSuite.providers.aws.resources.resources import Regions, AWSSimpleResources
+from ScoutSuite.providers.aws.facade.facade import AWSFacade
```
misg: Unused import.

ScoutSuite/providers/aws/services/awslambda/service.py
```diff
@@ -0,0 +1,30 @@
...
+from opinel.utils.aws import build_region_list
```
misg: Unused import.

ScoutSuite/providers/aws/services/ec2/snapshots.py
Aboisier: Yes, the rest of the info is relevant, as it is used in rules and is displayed in the report.

ScoutSuite/providers/aws/services/ec2/volumes.py
Aboisier: My previous answer applies to volumes as well.

ScoutSuite/providers/aws/resources/resources.py
```diff
@@ -0,0 +1,54 @@
...
+
+    async def fetch_all(self):
+        raw_resources = await self.get_resources_from_api()
+        for raw_resource in raw_resources:
+            name, resource = self.parse_resource(raw_resource)
+            self[name] = resource
+
+class AWSCompositeResources(CompositeResources, metaclass=abc.ABCMeta):
+
+    # TODO: get rid of the credentials.
```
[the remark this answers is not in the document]
Aboisier: Indeed.

ScoutSuite/providers/aws/resources/resources.py
Aboisier: Probably! But I actually use the credentials after all.

ScoutSuite/providers/aws/services/ec2/vpcs.py
Aboisier: Yea, I realized it's actually a case of diamond problem (if you were to call `super.fetch_all`, which parent's `fetch_all` should be called :|?)

misg: > Don't like the fact that `CompositeResources` inherits the `get_resources_from_api`, because it won't be used most of the time.

Not so sure about that, so far it could be used all the time since we have a call to `self.facade` in the `fetch_all` method of every classe that inherits from `CompositeResources`.
But surely we can postpone this conversation and wait for other cases where you would need to inherit from both `CompositeResources` and `SimpleResources`. :)

ScoutSuite/providers/aws/facade/utils.py
```diff
@@ -0,0 +1,26 @@
+from typing import Callable
+import boto3
+
+# TODO: Add docstrings
+class AWSFacadeUtils:
+    _clients = {}
+    
+    @staticmethod
+    def get_all_pages(service, region, paginator_name: str, response_key: str, **paginator_args):
```
misg: I find a bit weird to specify type hints for only half of the arguments, would it be possible to add them for `service` and `region` as well?

misg:
```suggestion
    def get_all_pages(service: str, region: str, paginator_name: str, response_key: str, **paginator_args):
```

ScoutSuite/providers/aws/resources/resources.py
```diff
@@ -0,0 +1,49 @@
...
+        self.scope = scope
+        self.facade = AWSFacade()
+
+
+class AWSCompositeResources(AWSResources, CompositeResources, metaclass=abc.ABCMeta):
+
+    """This class represents a collection of AWSResources. Classes extending AWSCompositeResources should \
+    define a "_children" attribute which consists of a list of tuples describing the children. The tuples \
+    are expected to respect the following format: (<child_class>, <child_name>). The child_name is used by \
+    indicates the name under which the child will be stored in the parent object.
```
misg: *is used to indicate

misg: Would it be worth to mention than `None` can be passed as `<child_name>` to indicate that the content of the child will be directly included in the parent (as result of the dict operation `update()`)?

misg: (Please disregard my previous comment if you're actually not doing this with AWS... Just thought about it because I'm doing so with Azure.)

ScoutSuite/providers/aws/services/ec2/instances.py
```diff
@@ -0,0 +1,31 @@
...
+
+class EC2Instances(AWSResources):
+    async def fetch_all(self, **kwargs):
+        raw_instances  = self.facade.ec2.get_instances(self.scope['region'], self.scope['vpc'])
+        for raw_instance in raw_instances:
+            name, resource = self._parse_instance(raw_instance)
+            self[name] = resource
+
+    def _parse_instance(self, raw_instance):
+        instance = {}
```
misg: I think we got an opportunity with this migration to be PEP 8 compliant and start defining dictionaries like the following:

```python
        instance = {
                'id' = raw_instance['InstanceId'],
                'reservation_id' = raw_instance['ReservationId'],
                'monitoring_enabled' = raw_instance['Monitoring']['State'] == 'enabled',
                'user_data' = self.facade.ec2.get_instance_user_data(self.scope['region'], id)
        }
```

ScoutSuite/providers/aws/services/ec2/securitygroups.py
```diff
@@ -0,0 +1,83 @@
...
+    icmp_message_types_dict = load_data('icmp_message_types.json', 'icmp_message_types')
+
+    async def fetch_all(self, **kwargs):
+        raw_security_groups = self.facade.ec2.get_security_groups(self.scope['region'], self.scope['vpc'])
+        for raw_security_groups in raw_security_groups:
+            name, resource = self._parse_security_group(raw_security_groups)
+            self[name] = resource
+
+    def _parse_security_group(self, raw_security_group):
+        security_group = {}
```
misg: Same comment as above.

ScoutSuite/providers/aws/facade/utils.py
misg: (I made a code suggestion just below.)

Aboisier: It's indeed weird, I guess I just forgot. 🙅‍♂️

ScoutSuite/providers/aws/resources/resources.py
Aboisier: Not sure this is a use case for AWS. Thanks for pointing it out tho!

ScoutSuite/providers/aws/services/ec2/instances.py
Aboisier: I talked about it with Rémi last Thursday, and we came to the conclusion it's probably a better idea to continue with the current implementation. There were a couple of reasons why, but here are the most important IMHO:

1. *Easier debugging*: let's say raw_instance['InstanceId'] fails because the dict does not have this key. Using the PEP 8 convention, we would not know exactly what caused the crash.
2. Sometimes, you want to add stuff iteratively. That's not possible with the PEP 8 convention unless you want to use a hybrid of the two notations, but then we might as well keep up with the current solution

PR #203, opened 2019-02-23
Feature/58 google compute engine findings

Remi05:
Added the checks for the 6 findings for Google Compute Engine defined by the CIS. The PR for the rules themselves can be found on the proprietary repo.

The newly implemented findings are the following:
- Ensure that instances are not configured to use the default service account with full access to all Cloud APIs. (CIS 4.1)
- Ensure "Block Project-wide SSH keys" enabled for VM instances. (CIS 4.2)
- Ensure oslogin is enabled for a Project. (CIS 4.3)
- Ensure 'Enable connecting to serial ports' is not enabled for VM Instance. (CIS 4.4)
- Ensure that IP forwarding is not enabled on Instances. (CIS 4.5)
- Ensure VM disks for critical VMs are encrypted with CustomerSupplied Encryption Keys. (CIS 4.6)

### DIFF
--- a/ScoutSuite/providers/gcp/services/computeengine.py
+++ b/ScoutSuite/providers/gcp/services/computeengine.py
@@ -210,8 +210,7 @@ class ComputeEngineConfig(GCPBaseConfig):
         return metadata.get('block-project-ssh-keys') == 'true'
 
     def _get_common_instance_metadata_dict(self, project_id):
-        request = self.api_client.projects().get(project=project_id)
-        project = request.execute()
+        project = self.api_client.projects().get(project=project_id).execute()
         return self._metadata_to_dict(project['commonInstanceMetadata'])
 
     def _is_oslogin_enabled(self, instance, project_id):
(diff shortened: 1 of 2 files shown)


ScoutSuite/providers/gcp/services/computeengine.py
```diff
@@ -188,3 +197,39 @@ def parse_firewalls(self, firewall, params):
...
+    def _metadata_to_dict(self, metadata):
+        return dict((item['key'], item['value']) for item in metadata['items']) if 'items' in metadata else {}
+
+    def _is_block_project_ssh_keys_enabled(self, instance):
+        metadata = self._metadata_to_dict(instance['metadata'])
+        return metadata.get('block-project-ssh-keys') == 'true'
+
+    def _get_common_instance_metadata_dict(self, project_id):
+        computeengine_client = gcp_connect_service(service='computeengine')
+        request = computeengine_client.projects().get(project=project_id)
```
Aboisier: I think you should be able to get an instance of the client by hsing

```py
params['api_client'].compute_engine()
```
Or something like that (not sure about the `compute_engine` part.

Remi05: Done!

j4v: Yep, you don't need to instantiate a client from the service object as it's passed down to it.

PR #204, opened 2019-02-24
Remove Opinel dependency

vifor2:
This PR is related to [this issue](https://github.com/nccgroup/ScoutSuite/issues/181).
It removes the references to the Opinel library since it is now deprecated, the proper files have been cloned and moved. Unused functions were removed and the whole been greatly cleaned up.


ScoutSuite/providers/aws/configs/regions.py
```diff
@@ -73,7 +76,7 @@ def __init__(self, service_metadata=None, thread_config=4):
                     continue
                 params = resource_metadata['params'] if 'params' in resource_metadata else {}
                 ignore_exceptions = True if 'no_exceptions' in resource_metadata and \
```
Remi05:
```suggestion
                 ignore_exceptions = resource_metadata.get('no_exceptions', False)
```

ScoutSuite/providers/aws/services/cloudwatch.py
```diff
@@ -28,12 +28,11 @@ def parse_alarm(self, global_params, region, alarm):
         alarm['name'] = alarm.pop('AlarmName')
         # Drop some data
         for k in ['AlarmConfigurationUpdatedTimestamp', 'StateReason', 'StateReasonData', 'StateUpdatedTimestamp']:
-            foo = alarm.pop(k) if k in alarm else None
+            alarm.pop(k) if k in alarm else None
```
Remi05:
```suggestion
             if k in alarm:
                 alarm.pop(k)
```
Since there is no actual assignment done, I believe it is cleaner to use a normal if statement.

ScoutSuite/providers/aws/services/iam.py
```diff
@@ -152,18 +154,16 @@ def parse_groups(self, group, params):
...
         """
         Parse a single IAM policy and fetch additional information
         """
         api_client = params['api_client']
-        policy = {}
-        policy['name'] = fetched_policy.pop('PolicyName')
-        policy['id'] = fetched_policy.pop('PolicyId')
-        policy['arn'] = fetched_policy.pop('Arn')
+        policy = {'name': fetched_policy.pop('PolicyName'), 'id': fetched_policy.pop('PolicyId'),
+                  'arn': fetched_policy.pop('Arn')}
```
Remi05: I would keep this as it was for consistency since the `parse_...()` methods in the other services are all written like this.

ScoutSuite/providers/aws/services/iam.py
```diff
@@ -210,30 +210,24 @@ def fetch_password_policy(self, credentials):
...
-                self.password_policy['RequireUppercaseCharacters'] = False
-                self.password_policy['RequireLowercaseCharacters'] = False
-                self.password_policy['RequireNumbers'] = False
-                self.password_policy['RequireSymbols'] = False
-                self.password_policy['PasswordReusePrevention'] = False
-                self.password_policy['ExpirePasswords'] = False
+                self.password_policy = {'MinimumPasswordLength': '1', 'RequireUppercaseCharacters': False,
+                                        'RequireLowercaseCharacters': False, 'RequireNumbers': False,
+                                        'RequireSymbols': False, 'PasswordReusePrevention': False,
+                                        'ExpirePasswords': False}
```
Remi05: Same here, I would keep it as it was. Or if we ever decide to change all the services, I would at least put the different entries on new lines as it is much more readable.

ScoutSuite/providers/aws/services/rds.py
```diff
@@ -66,19 +64,21 @@ def parse_snapshot(self, global_params, region, dbs):
         for attribute in attributes:
             snapshot[attribute] = dbs[attribute] if attribute in dbs else None
         api_client = api_clients[region]
-        attributes = api_client.describe_db_snapshot_attributes(DBSnapshotIdentifier = snapshot_id)['DBSnapshotAttributesResult']
+        attributes = api_client.describe_db_snapshot_attributes(DBSnapshotIdentifier=snapshot_id)[
+            'DBSnapshotAttributesResult']
```
Remi05: Do we really want to enforce the maximum line length all the time? I'm not a big fan of splitting a line like this.

ScoutSuite/providers/aws/services/rds.py
```diff
@@ -138,11 +136,10 @@ def get_security_groups_info(rds_client, region_info):
     for group in groups:
         region_info['vpcs'][ec2_classic]['security_groups'][group['DBSecurityGroupName']] = parse_security_group(group)
 
+
 def parse_security_group(group):
-    security_group = {}
-    security_group['name'] = group['DBSecurityGroupName']
-    security_group['description'] = group['DBSecurityGroupDescription']
-    security_group['ec2_groups'] = {}
+    security_group = {'name': group['DBSecurityGroupName'], 'description': group['DBSecurityGroupDescription'],
```
Remi05: Once more here, we should keep it consistent across all services (aka. how it was before).

ScoutSuite/providers/aws/services/redshift.py
```diff
@@ -44,33 +43,30 @@ def parse_parameter_group(self, global_params, region, parameter_group):
...
         parameter_group['parameters'] = {}
         api_client = api_clients[region]
-        parameters = handle_truncated_response(api_client.describe_cluster_parameters, {'ParameterGroupName': pg_name}, ['Parameters'])['Parameters']
+        parameters = handle_truncated_response(api_client.describe_cluster_parameters, {'ParameterGroupName': pg_name},
+                                               ['Parameters'])['Parameters']
         for parameter in parameters:
-            param = {}
-            param['value'] = parameter['ParameterValue']
-            param['source'] = parameter['Source']
+            param = {'value': parameter['ParameterValue'], 'source': parameter['Source']}
```
Remi05: Same.

ScoutSuite/providers/aws/services/s3.py
```diff
@@ -137,6 +133,7 @@ def update_iam_permissions(s3_info, bucket_name, iam_entity, allowed_iam_entity,
 
 def update_bucket_permissions(s3_info, iam_info, action, iam_entity, allowed_iam_entity, full_path, policy_type,
                               policy_name):
+    global policy
```
Remi05: Why do we need a `global` variable?

ScoutSuite/providers/aws/services/s3.py
```diff
@@ -167,11 +163,7 @@ def update_bucket_permissions(s3_info, iam_info, action, iam_entity, allowed_iam
 
 
 def init_s3_permissions():
-    permissions = {}
-    permissions['read'] = False
-    permissions['write'] = False
-    permissions['read_acp'] = False
-    permissions['write_acp'] = False
+    permissions = {'read': False, 'write': False, 'read_acp': False, 'write_acp': False}
```
Remi05: Same here too.

ScoutSuite/utils.py
```diff
@@ -90,4 +87,26 @@ def is_throttled(e):
     :param e:                           Exception raised
     :return:                            True if it's a throttling exception else False
     """
-    return True if  (hasattr(e, 'response') and 'Error' in e.response and e.response['Error']['Code'] in [ 'Throttling', 'RequestLimitExceeded', 'ThrottlingException' ]) else False
+    return True if (hasattr(e, 'response')
```
Remi05:
```suggestion
    return (hasattr(e, 'response') and 'Error' in e.response
                and e.response['Error']['Code'] in ['Throttling', 'RequestLimitExceeded', 'ThrottlingException'])
```
If  you apply the suggestion, make sure that not just the first line is replaced.

ScoutSuite/providers/aws/configs/regions.py
Remi05: Also, make sure that not just the first line is replaced, I didn't mange to select the whole statement at once.

ScoutSuite/providers/aws/services/iam.py
zer0x64: I would personnaly disagree, as it would be better to do it the "right" way(PEP8 compliant) while we're working on it and eventually open a PR to change the remaining ones.

ScoutSuite/providers/aws/services/rds.py
vifor2: This way is much cleaner and PyCharm suggests it as a correction though...

ScoutSuite/providers/aws/services/iam.py
zer0x64: Same as before, but I would agree that it would be better to split it 1-line per arguments

ScoutSuite/providers/aws/services/rds.py
zer0x64: PEP8 ;)

zer0x64: One of the high points of respecting coding styles is to prevent conflicts. If everyone follows the coding styles, git won't flag a bunch of conflicts where the same line is changed the same way, but with different spacing and stuff

ScoutSuite/providers/aws/services/rds.py
zer0x64: Same asnwer as above

ScoutSuite/providers/aws/services/redshift.py
zer0x64: Same

ScoutSuite/providers/aws/services/s3.py
zer0x64: Same

ScoutSuite/providers/aws/services/iam.py
vifor2: Yeah, I agree with @zer0x64 for this kind of issue, as I've written on another comment.

ScoutSuite/providers/aws/services/s3.py
vifor2: To prevent the warning `Local variable 'policy' might be referenced before assignment` thrown by PyCharm.

zer0x64:
Quick note: In conditions.py I have put two different import inside functions to fix a circular depedency. It may not be the cleanest thing to do, but I don't see how to fix it otherwise. Suggestions are welcome!

ScoutSuite/providers/aws/services/s3.py
zer0x64: A better way to do this is to actually assign it to an empty dictionary at the beginning of the function. Pushing this now

j4v:
> circular depedency

What's the dependency? Maybe an `import as ...` could do the trick?

zer0x64:
> > circular depedency
> 
> What's the dependency? Maybe an `import as ...` could do the trick?

This only happens in the tests, but not when running the tool as normal.  I don't really see how a `import as` could fix this? I know replacing the "from ....browser import..." to a "import ...browser" does not fix it.

image

ScoutSuite/providers/aws/services/iam.py
Remi05: @Aboisier @misg Do you feel like we should change these for now and the rest later or change all of them at the same time later on?

ScoutSuite/core/console.py
```diff
@@ -35,74 +32,68 @@ def configPrintException(enable):
...
+def prompt(test_input=None):
     """
     Prompt function that works for Python2 and Python3
 
     :param test_input:                  Value to be returned when testing
 
     :return:                            Value typed by user (or passed in argument when testing)
     """
-    if test_input != None:
+    if not test_input:
```
Aboisier: Was that change intentional? It seems like it's reverting the condition.

ScoutSuite/core/console.py
```diff
@@ -111,45 +102,33 @@ def prompt_4_mfa_code(activate = False, input = None):
...
-            mfa_code[5]
-            break
-        except:
-            printError('Error: your MFA code must only consist of digits and be at least 6 characters long.')
+        mfa_code = prompt_value(prompt_string, no_confirm=True, test_input=test_input)
+        if mfa_code == 'q':
+            return mfa_code
+        if not re_mfa_code.match():
+            print_error('Error: your MFA code must only consist of digits and be at least 6 characters long.')
+        break
```
Aboisier: I really prefer the previous implementation, it was l33t3r. (jokes)

ScoutSuite/core/console.py
```diff
@@ -165,92 +144,94 @@ def prompt_4_value(question, choices = None, default = None, display_choices = T
...
 
 
-def prompt_4_yes_no(question, input = None):
+def prompt_yes_no(question, test_input=None):
     """
     Prompt for a yes/no or y/n answer
                                         .
     :param question:                    Question to be asked
-    :param input:                       Used for unit testing
+    :param test_input:                       Used for unit testing
```
Aboisier:
```suggestion
    :param test_input:                   Used for unit testing
```

ScoutSuite/core/rule_definition.py
```diff
@@ -72,15 +72,15 @@ def load(self):
...
         if not file_name_valid:
-            printError('Error: could not find %s' % self.file_name)
+            print_error('Error: could not find %s' % self.file_name)
         else:
             try:
                 with open(self.file_path, 'rt') as f:
                     self.string_definition = f.read()
                     self.load_from_string_definition()
             except Exception as e:
                 # printException(e)
```
Aboisier:
```suggestion
```

ScoutSuite/core/ruleset.py
```diff
@@ -85,13 +85,13 @@ def load(self, rule_type, quiet=False):
                             self.handle_rule_versions(filename, rule_type, rule)
             except Exception as e:
                 # printException(e)
```
Aboisier:
```suggestion
```

ScoutSuite/core/rule.py
```diff
@@ -140,4 +139,4 @@ def set_definition(self, rule_definitions, attributes = None, ip_ranges = None,
                 setattr(self, 'key', '%s-%s' % (self.key, self.key_suffix))
         except Exception as e:
             # printException(e)
```
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/aws.py
```diff
@@ -146,4 +142,6 @@ def is_throttled(e):
     :param e:                           Exception raised
     :return:                            True if it's a throttling exception else False
     """
-    return True if  (hasattr(e, 'response') and 'Error' in e.response and e.response['Error']['Code'] in [ 'Throttling', 'RequestLimitExceeded', 'ThrottlingException', 'TooManyRequestsException' ]) else False
+    return True if (hasattr(e, 'response') and 'Error' in e.response and e.response['Error']['Code'] in
+                    ['Throttling', 'RequestLimitExceeded', 'ThrottlingException', 'TooManyRequestsException']) \
+        else False
```
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/aws.py
```diff
@@ -146,4 +142,6 @@ def is_throttled(e):
     :param e:                           Exception raised
     :return:                            True if it's a throttling exception else False
     """
-    return True if  (hasattr(e, 'response') and 'Error' in e.response and e.response['Error']['Code'] in [ 'Throttling', 'RequestLimitExceeded', 'ThrottlingException', 'TooManyRequestsException' ]) else False
+    return True if (hasattr(e, 'response') and 'Error' in e.response and e.response['Error']['Code'] in
```
Aboisier:
```suggestion
    return (hasattr(e, 'response') and 'Error' in e.response and e.response['Error']['Code'] in
```

ScoutSuite/providers/aws/aws.py
```diff
@@ -146,4 +142,6 @@ def is_throttled(e):
...
+                    ['Throttling', 'RequestLimitExceeded', 'ThrottlingException', 'TooManyRequestsException']) \
```
Aboisier:
```suggestion
                    ['Throttling', 'RequestLimitExceeded', 'ThrottlingException', 'TooManyRequestsException'])
```

ScoutSuite/providers/aws/services/iam.py
Aboisier: This PR is already painful enough. Let's change this iteratively as we migrate the services to the new architecture. What do you guys think?

ScoutSuite/providers/aws/services/route53.py
```diff
@@ -66,9 +60,10 @@ def parse_hosted_zones(self, hosted_zone, params):
...
-        record_sets = handle_truncated_response(api_client.list_resource_record_sets, {'HostedZoneId': hosted_zone_id}, ['ResourceRecordSets'])
+        record_sets = handle_truncated_response(api_client.list_resource_record_sets, {'HostedZoneId': hosted_zone_id},
+                                                ['ResourceRecordSets'])
         hosted_zone.update(record_sets)
-        #print(str(record_sets))
-        #record_sets = api_client.list_resource_record_sets()
-        #hosted_zone['RecordSets'] = record_sets['Resourc']
+        # print(str(record_sets))
+        # record_sets = api_client.list_resource_record_sets()
+        # hosted_zone['RecordSets'] = record_sets['Resourc']
```
Aboisier: Might as well delete it?

ScoutSuite/utils.py
```diff
@@ -45,49 +35,28 @@
...
     """
-    for key in keys:
-        #dst[no_camel(key)] = src[key] if key in src else None
-        dst[key] = src[key] if key in src else None
+    if not str(key) in dictionary:
+        dictionary[str(key)] = init
+        manage_dictionary(dictionary, key, init)
+        if callback:
+            callback(dictionary[key])
+    return dictionary
```
Aboisier: I suggest using a guard clause in order to reduce nesting.

```py
def manage_dictionary(dictionary, key, init, callback=None):
    """
    :param dictionary:
    :param key:
    :param init:
    :param callback:
    :return:
    """

    if str(key) in dictionary:
        return dictionary

    dictionary[str(key)] = init
    manage_dictionary(dictionary, key, init)
    if callback:
        callback(dictionary[key])
    return dictionary
```

j4v:
Great job guys :muscle:

Issue #101, opened 2019-01-16
Public bucket objects not identified


allisonis:
Currently the checks in buckets.py are looking solely at bucket level permissions. Since object level acl's can differ from bucket level permissions, publicly accessible objects are not identified. In order to determine if objects are publicly accessible acl's could be checked as well for the allUsers entity.

j4v:
You are correct, this is also the case for AWS S3 bucket objects. The reason objects aren't being checked is that buckets may have thousands++ objects: checking each one of them could potentially make scans run forever.

Finding a solution to this is indeed in the pipeline, will keep this issue open for tracking.

j4v:
Closing this and marking as `wontfix` as analyzing all of the objects in a bucket doesn't scale.

Issue #57, opened 2018-12-14
new finding - GKE Default Service Account In Use


j4v:
Finding for the GCP GKE service.

Check this is contained in the cluster details:
```
node_config {
  […]
  service_account: "default"
}
```

PR #205, opened 2019-02-25
Feature/57 gcp cloud storage findings prop

Aboisier:
This PR is part of [#57](https://github.com/nccgroup/ScoutSuite-Proprietary/issues/57) on the proprietary side. CIS 5.1 and 5.3 were already implemented, I added their rationales and CIS numbers. With the current API, CIS 5.2 would require hours to complete, since it requires potentially getting millions of bucket objects. See #101 .


Remi05:
Looks good to me!

PR #206, opened 2019-02-26
Feature/91 cloudformation findings

vifor2:
PR in response to [this issue](https://github.com/nccgroup/ScoutSuite-Proprietary/issues/97).
Please also review the [private part](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/101) of this PR.

What's new:
- Now fetching the information for the 2 new rules.
- These informations are view-able and highlighted when clicking on the rule, see GIF (view-able on both public and private versions, but the highlighting is only on the private version since the rule is on the private repository).

ezgif-3-cd4e07a241ad

### DIFF
--- a/ScoutSuite/providers/aws/rules/findings/cloudformation-stack-with-role.json
+++ b/ScoutSuite/providers/aws/rules/findings/cloudformation-stack-with-role.json
@@ -1,5 +1,5 @@
 {
-    "description": "Role passed to stack",
+    "description": "Stack should not have a role",
     "rationale": "Passing a role to CloudFormation stacks may result in privilege escalation because IAM users with privileges within the CloudFormation scope implicitly inherit the stack's role's permissions.",
     "path": "cloudformation.regions.id.stacks.id",
     "dashboard_name": "Stacks",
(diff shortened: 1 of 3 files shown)


ScoutSuite/providers/aws/rules/findings/cloudformation-stack-with-role.json
```diff
@@ -1,5 +1,5 @@
 {
-    "description": "Role passed to stack",
+    "description": "Stack should not have a role",
```
j4v: I'd leave it as it was. It's not so much that passing a role to a stack is bad (it's not), but that the user should be knowledgeable of it, and ensure the IAM privileges given to that role follow the principle of least privilege.

ScoutSuite/providers/aws/rules/findings/cloudformation-stack-with-role.json
```diff
@@ -1,5 +1,5 @@
...
     "rationale": "Passing a role to CloudFormation stacks may result in privilege escalation because IAM users with privileges within the CloudFormation scope implicitly inherit the stack's role's permissions.",
```
j4v: You could instead amend the rationale:
```
Passing a role to CloudFormation stacks may result in privilege escalation because IAM users with privileges within the CloudFormation scope implicitly inherit the stack's role's permissions. Consequently, it should be ensured that the IAM privileges assigned to the stack's role follow the principle of least privilege.
```

Issue #129, opened 2019-01-30
Menu highlight is incorrect when clicking the Scout Suite logo


xlash:
Steps to reproduce :

1- Click on any menu header, in this example, Storage
2- Click on Scout2 Logo to go back to general findings
3- Select any finding categories
4- "Storage" is still the selected menu

Tested on Chrome 71.0.3578.98 (Official Build) (64-bit)
Report generated with Scout Suite version 4.0.6

![image](https://user-images.githubusercontent.com/50488/51994844-9de52f00-247f-11e9-996a-2fddb29cb1a4.png)

j4v:
Thanks for opening this. 

You might want to check the `develop` branch to confirm whether this issue is present in the latest version.

xlash:
Reproduce with develop branch. 

![image](https://user-images.githubusercontent.com/50488/52002839-ff61c980-2490-11e9-8fa2-60b79ae83d25.png)
Report generated with Scout Suite version 4.0.6

j4v:
@vifor2 can you have a look at this please?

Aboisier:
Closing, fixed in #208 . Thanks for reporting this bug, @xlash!

PR #208, opened 2019-02-27
Bugfix/129 menu highlight

Aboisier:
This PR fixes the broken menu highlighting, as described in #129 .

Sorry for the parasites diffs in `scoutsuite.js`, I accidentally formatted the whole file. It'll be done once and for all at least.


j4v:
:fire:

PR #210, opened 2019-02-28
bugfix/Single profile instead of array

zer0x64:
Will require some testing on the profiles, as I don't have the setup to verify AWS profiles.

This small PR aims to make the --profile command line argument into a single variable instead of an array which we only use the first argument.

### DIFF
--- a/ScoutSuite/__main__.py
+++ b/ScoutSuite/__main__.py
@@ -37,7 +37,7 @@ def main(args=None):
 
     # Create a cloud provider object
     cloud_provider = get_provider(provider=args.get('provider'),
-                                  profile=args.get('profile')[0] if args.get('profile') else None,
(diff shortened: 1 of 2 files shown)


j4v:
Tested, looks good. These small changes are slowly but surely improving the code quality!

ScoutSuite/__main__.py
```diff
@@ -124,7 +124,7 @@ def main(args=None):
 
     # Handle exceptions
     try:
-        exceptions = RuleExceptions(args.get('profile')[0] if args.get('profile') else None, args.get('exceptions')[0])
+        exceptions = RuleExceptions(args.get('profile'), args.get('exceptions')[0])
```
j4v: Maybe the same could/should be done for `exceptions`?

PR #214, opened 2019-03-03
Refactoring/azure/sqldatabase

Aboisier:
This PR aims to migrate Azure SQL Database service to the new architecture, as part of #183. The work that has been done was already in #197 (you should take a look at it if you want more context with the conversation there) but because of some wrong git manipulations, we had to start again with a new branch hence this new PR.


ScoutSuite/providers/azure/resources/utils.py
```diff
@@ -0,0 +1,14 @@
+from hashlib import sha1
+
+
+def get_non_provider_id(name):
+        """
+        Not all resources have an ID and some services allow the use of "." in names, which break's Scout2's
```
Aboisier:
```suggestion
        Not all resources have an ID and some services allow the use of "." in names, which breaks Scout's
```

ScoutSuite/providers/azure/resources/sql/transparent_data_encryptions.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
Aboisier: Is this necessary? I think the python interpreter now uses utf-8 implicitly.

ScoutSuite/providers/azure/resources/sql/servers.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+
+from azure.mgmt.sql import SqlManagementClient
+
+from ..resources import AzureCompositeResources
+from ScoutSuite.providers.azure.utils import get_resource_group_name
+
+from .databases import Databases
+from .server_azure_ad_administrators import ServerAzureAdAdministrators
+from ..utils import get_non_provider_id
```
Aboisier: I believe we've been mostly using absolute paths. I find it's more verbose, especially for very vague names like utils.

Aboisier:
It seems like a test is failing. Otherwise, I approve, looks good!

ScoutSuite/providers/azure/resources/sql/transparent_data_encryptions.py
j4v: It's the default encoding for python 3 (not 2, which is raw data). Seeing how the refactoring is removing support for 2.x, we can leave this out I guess.

ScoutSuite/providers/azure/resources/utils.py
```diff
@@ -0,0 +1,14 @@
...
+        recursion scheme if name is used as an ID. Use SHA1(name) instead.
+
+        :param name:                    Name of the resource to
+        :return:                        SHA1(name)
+        """
+        m = sha1()
+        m.update(name.encode('utf-8'))
+        return m.hexdigest()
```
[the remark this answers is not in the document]
j4v: Yeah, please use verbose names for variables, let's not repeat past mistakes. Bytes are cheap!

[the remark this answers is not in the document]
Aboisier: This is used by AWS, Azure and GCP. We could move it out of the azure folder.

ScoutSuite/providers/azure/resources/sql/servers.py
misg: I agree for `..utils`. Although, I would suggest to use relative paths to refer to files of a same directory and use absolute paths for all other cases (when you need to import a file that is at least one level above in the file tree). That way, we're still taking advantages of Python package system to write imports like `from .database_blob_auditing_policies import DatabaseBlobAuditingPolicies` instead of `from ScoutSuite.providers.azure.resources.sqldatabase.database_blob_auditing_policies import DatabaseBlobAuditingPolicies`.

Issue #9, opened 2018-08-27
Leverage boto3 for authentication


j4v:
The current implementation for authentication could/should rely solely on boto3 and not on any custom code, AKA remnants of Opinel.

It should be validated that this wouldn't create any issues with MFA.

vifor2:
Current branch : ­[refactoring/9-leverage-boto3-for-authentication](https://github.com/nccgroup/ScoutSuite/tree/refactoring/9-leverage-boto3-for-authentication)

TODO : Replace the contents of `crendentials.py` with an implementation of authentication via boto3, reading credentials from a CSV file will be dropped by doing so.

vifor2:
To make sure I don't forget stuff I like to write it down, and since I'm not working on this issue alone I figure I'll write my current thoughts here :
- Other than for the tests, the only functions from `credentials.py` being called is `read_creds`, which greatly simplifies our work since by a first look at  `credentials.py` we could've thought that there is much more to do since there are 15 functions defined over 533 lines.
![image](https://user-images.githubusercontent.com/17322874/53699608-3856cc00-3db8-11e9-8793-00328ebe5248.png)
- `read_creds` : Inputs : profile_name, ~~csv_file~~, mfa_serial_arg, mfa_code, force_init. Outputs : dict containing : AccessKeyId {str}, SecretAccessKey {str}, SessionToken {str}, Expiration {datetime}.

j4v:
You're right that only `read_creds` is explicitly called, but most of the functions in `credentials.py` are then used by `read_creds` (`read_creds` implements the logic, the other functions do the actual IAM work).

This said I'd think they can all be scraped for a much simpler implementation:
```python
# in this case a profile is explicitly stated
if args.profile:
    session = boto3.Session(profile_name='<foo>')
# otherwise the boto logic is used, which should work natively for EC2, ECS, Lambda, etc
# logic defined here https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#configuring-credentials
else: 
       session = boto3.Session()       
```

The `session` can then be reused (instead of "credentials" as currently done) in `connect_service` (https://github.com/nccgroup/ScoutSuite/blob/1b4b502822e1b4f9bdc5cf4bfd394a60202e9461/opinel/utils/aws.py#L31):

```python
# Any clients created from the above session will use those credentials
s3_client = session.client('s3')
```

vifor2:
### Update
[...]

PR #215, opened 2019-03-04
Refactoring - Leverage boto3 for Authentication

vifor2:
PR in response to [this issue](https://github.com/nccgroup/ScoutSuite/issues/9).
Currently WIP till further feedback/approval.

What has been done : 
- Got rid of `credentials.py` which came from Opinel and by the same move we also got rid of `Requests`
- Authentification is now made with boto3's `Session()` instead
- Removed the functionality of using Scout Suite with a CSV (since if you can create one you can also simply modify the `.aws/credentials` file, which is the standard/documented AWS method)
- Removed the prompt for MFA code since it was only called on one specific case (with roles) and it doesn't really make it easier to use Scout Suite


ScoutSuite/__main__.py
```diff
@@ -165,6 +162,7 @@ def main(args=None):
 
 
 def generate_report_name(provider_code, args):
+    report_file_name = {}
```
Remi05: Is there a reason to assign a dictionary to `report_file_name`? Two lines lower, a string is assigned to it. 😕

ScoutSuite/core/console.py
```diff
@@ -13,7 +13,7 @@
 
 mfa_serial_format = r'^arn:aws:iam::\d+:mfa/[a-zA-Z0-9\+=,.@_-]+$'
 re_mfa_serial_format = re.compile(mfa_serial_format)
-re_mfa_code = re.compile(r'^\d{6}\d*$')
+re_mfa_code = r'^\d{6}\d*$'
```
Remi05:
```suggestion
re_mfa_code_pattern = r'^\d{6}\d*$'
```
I believe this would be a clearer variable name.

ScoutSuite/providers/base/provider.py
```diff
@@ -112,21 +110,18 @@ def _load_metadata(self):
...
     def _update_last_run(self, current_time, ruleset):
-        last_run = {}
-        last_run['time'] = current_time.strftime("%Y-%m-%d %H:%M:%S%z")
-        last_run['cmd'] = ' '.join(sys.argv)
-        last_run['version'] = scout2_version
-        last_run['ruleset_name'] = ruleset.name
-        last_run['ruleset_about'] = ruleset.about
-        last_run['summary'] = {}
+        last_run = {'time': current_time.strftime("%Y-%m-%d %H:%M:%S%z"), 'cmd': ' '.join(sys.argv),
+                    'version': scout2_version, 'ruleset_name': ruleset.name, 'ruleset_about': ruleset.about,
```
Remi05: Could we put keys on different lines? I find it more readable when there are many keys or when the values are somewhat long.

tests/test-scoutsuite.py
```diff
@@ -69,7 +67,8 @@ def test_scout_suite_cis_ruleset_run(self):
 #    # Make sure that ScoutSuite's check-s3-acl option does not crash
 #    #
```
Remi05: Do we still need that commented code?

ScoutSuite/core/console.py
Remi05: Make sure to update the lines where it is used as well if you apply the suggestion 😉

tests/test-scoutsuite.py
vifor2: Don't know lol, I'll delete it

ScoutSuite/providers/base/provider.py
vifor2: Hmm sorry, but I prefer it this way, find me someone else in the team who prefers the long way and I'll change it 😉

j4v:
Made some changes & tested for:
- Named profile
- No named profile, which reverted to `default` configured profile
- No named profile or `default` configured profile, which reverted to profile configured in environment variables
- GCP/Azure to test integration

README should be updated. It would also be worthwhile to link to https://aws.amazon.com/premiumsupport/knowledge-center/authenticate-mfa-cli/ for MFA configuration.

PR #218, opened 2019-03-04
4.3.0 release

Aboisier:
PR for the 4.3.0 release.


zer0x64:
Shouldn't we add the Python2 deprecation warning in the README?

zer0x64:
Packaging on Python3.7 is working(tested on AWS and azure). Package for Python2.7 is broken on AWS(not on azure):

ScoutSuite/providers/aws/configs/services.py", line 56, in __init__
    super().__init__(metadata, thread_config)
TypeError: super() takes at least 1 argument (0 given)

Do we fix it or something else?

zer0x64:
> �
> 
> > Packaging on Python3.7 is working(tested on AWS and azure). Package for Python2.7 is broken on AWS(not on azure):
> > ScoutSuite/providers/aws/configs/services.py", line 56, in **init**
> > super().**init**(metadata, thread_config)
> > TypeError: super() takes at least 1 argument (0 given)
> > Do we fix it or something else?
> 
> We should probably fix it. Do you have an idea of what the cause is? I looked at the git history, I don't see what could have caused this problem. Was it even working before?

Fixed in https://github.com/nccgroup/ScoutSuite/pull/228

PR #223, opened 2019-03-05
Updated README following PR #215

vifor2:
Update README.md following [this PR](https://github.com/nccgroup/ScoutSuite/pull/215).

- Removed the mentions that Python 2.7 & 3.4 are supported since we will be getting rid of them for the next release.
- Now telling the user to configure MFA themselves following @j4v's changes.
- No longer telling the user they can identify with a CSV file since it has been removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,6 @@ permissions:
 If the usage of multi-factor authentication (MFA) is enabled, please consult [this page](https://aws.amazon.com/premiumsupport/knowledge-center/authenticate-mfa-cli/) in order to 
 configure MFA authentication through the management of session tokens.
 
-
 #### Google Cloud Platform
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,8 +61,6 @@ may result in the process being killed.
 
 Scout Suite is written in Python and supports the following versions:
 
--   2.7
--   3.4
```
zer0x64: We'll need this PR merged before the release, so add a warning about the future deprecation or Python2.7-3.4 instead of removing it now

j4v:
> Not sure how to deal with the "Python2 will be deprecated" in the current release, but other than that it's good!

I'd say update the readme with a warning and make it clear in the release notes. All of the refactoring will likely be released under a `5.0.0` number, so we can make a big deal of all the changes.

PR #224, opened 2019-03-05
refactoring/async-support

misg:
This PR aims to improve async support for the current refactoring (#183).
Basically, every API call to a cloud provider (made through boto for AWS or azure-mgmt for Azure, for example) done in ours facades should now
be wrapped around a lambda and run asynchronously using the helper `run_concurrently` defined in `providers/utils.py`.
Doing so, every API call to fetch external cloud ressources will be run in a dedicated thread pool (newly defined in `Scout.py`, as
an internal member of the global event loop), making all the fetching part of Scout Suite really concurrent and asynchronous.
Please look at this PR in order to define really async facades and ressources for the next new migrated services.


ScoutSuite/providers/aws/facade/utils.py
```diff
@@ -1,26 +1,32 @@
...
+
+        # Getting all pages from a paginator requires API calls so we need to do it concurrently:
+        return await run_concurrently(lambda: AWSFacadeUtils._get_all_pages_from_paginator(paginator, response_key))
 
     @staticmethod
-    def _get_from_all_pages(pages: [], key:str):
+    def _get_all_pages_from_paginator(paginator, key):
         resources = []
-        for page in pages:
+        # There's an API call hidden behind each iteration:
```
Aboisier: sneaky sneaky boto

Aboisier:
Here are the execution times for before and after the refactoring for different services. The new implementation is slightly slower, but it doesn't seem like it's really significant (it's not a whole order of magnitude slower, it's more of a teeny constant overhead).

image

j4v:
> Here are the execution times for before and after the refactoring for different services.

Against what are you running this? I'm asking because as we've seen in the past execution time depends greatly of how many resources are in the account. So let's say you run threading vs coroutines on a small account, I wouldn't expect much difference. Might be worth deploying a *large* amount of templates (e.g. choose 10 from https://github.com/awslabs/aws-cloudformation-templates, and deploy 10 times each template) and see what gives. I can do the deployment if you'd like.

There's also a difference between how fast the tool runs, and how fast it makes API calls. As we've seen for large accounts the rule engine was taking a lot of time - that's the type of thing we have to look to optimize. Because API calls can/are throttled by the cloud providers, and in many cases we want to be careful about how many we make*.

* This is important because for example with AWS the API throttling is account-wide. So let's say you're on an engagement with a client that has a very large account, with a complete CI/CD pipeline. You come in with Scout in hand and think "hey, I'll just run the tool at full throttle and finish the scan quickly"... Ensues hitting the account's API limit, freezing all of the internal operations (monitoring, deployment, CloudFormation updates, etc.). When an angry client calls you because you broke their internet, it's not a good day.

Aboisier:
> > Here are the execution times for before and after the refactoring for different services.
> 
> Against what are you running this? I'm asking because as we've seen in the past execution time depends greatly of how many resources are in the account. So let's say you run threading vs coroutines on a small account, I wouldn't expect much difference. Might be worth deploying a _large_ amount of templates (e.g. choose 10 from https://github.com/awslabs/aws-cloudformation-templates, and deploy 10 times each template) and see what gives. I can do the deployment if you'd like.
>
> There's also a difference between how fast the tool runs, and how fast it makes API calls. As we've seen for large accounts the rule engine was taking a lot of time - that's the type of thing we have to look to optimize. Because API calls can/are throttled by the cloud providers, and in many cases we want to be careful about how many we make*.

I understand your concern. @misg and I spent enough time tracing the execution of a scan that we can say we're pretty positive the number of api calls did not change significantly. The frequency of the api calls can still be modulated by the number of threads in the thread pool (I know you discussed this with @misg , I'm just putting this here for the others).

I can't really test with a significant amount of resources, it seems like my account has a quota of 20 VMs. I still am getting similar execution times (around 17s with the previous implementation, and  22s with the new one).

misg:
> Here are the execution times for before and after the refactoring for different services.
> 
> Against what are you running this?

Nothing crazy, it was just against one single VM with some instances, snapshots and vpcs (for a total of around 40 resources).

> I'm asking because as we've seen in the past execution time depends greatly of how many resources > are in the account. So let's say you run threading vs coroutines on a small account, I wouldn't expect much difference. Might be worth deploying a _large_ amount of templates (e.g. choose 10 from https://github.com/awslabs/aws-cloudformation-templates, and deploy 10 times each template) and see what gives. I can do the deployment if you'd like.

Indeed, it would be interesting to see if there's any difference at "large scale". I believe Antoine couldn't really create that much of resources, would be sweet of you if you could indeed help with the deployment, thanks! 

> There's also a difference between how fast the tool runs, and how fast it makes API calls. As we've seen for large accounts the rule engine was taking a lot of time - that's the type of thing we have to look to optimize.

To be honest, I don't think we will have enough time to look into the rule engine (but we'll see)... Our current refactoring primarily aims to make the fetching&parsing part of Scout Suite well designed and really homogeneous for all the cloud providers in order to increase a lot the maintenability, readability and extensibility of the software.

> Because API calls can/are throttled by the cloud providers, and in many cases we want to be careful about how many we make*.
> 
> *This is important because for example with AWS the API throttling is account-wide...

If I recall correctly, you can control how fast the API calls are made in the current version of Scout Suite by simply adjusting the number of threads used during its execution (through the CLI argument `thread-config`).
Our refactoring won't change that. We may have not been very clear on the topic of Python coroutines but there's no opposition to be made between coroutines and threads, actually. They are basically not the same thing and we're not replacing one by the other.
Python coroutines just allow us to use threads in a more elegant, transparent and maintenable way. That's it. We're still using a pool of threads (see [here](https://github.com/nccgroup/ScoutSuite/blob/aab240c2deb020b3ad7c6fb8dc9feb99aa3d8e29/Scout.py#L14) and [there](https://github.com/nccgroup/ScoutSuite/blob/aab240c2deb020b3ad7c6fb8dc9feb99aa3d8e29/ScoutSuite/providers/utils.py#L18) for example) under the hood to fetch and parse cloud resources and you can still adjust the number of threads used in order to throttle the number of API calls made.

Aboisier:
> Indeed, it would be interesting to see if there's any difference at "large scale". I believe Antoine couldn't really create that much of resources, would be sweet of you if you could indeed help with the deployment, thanks!

Indeed, it seems like I have a quota of 20 instances. Testing with a total of 123 resources (images, snapshots, instances, vpcs, etc.), I was still getting a decent performance (17s before the refactoring vs 22s after the refactoring).

j4v:
>  I don't think we will have enough time to look into the rule engine

Understood. I think there are 2 quick wins (remove all the deepcopy as it's likely unnecessary and makes things very slow & simply parallelize the execution) which would make significant improvements.

> Our refactoring won't change that.

Ah, I had not understood that. Great then. It's also likely why the execution time cannot drastically improve, because in the background what's happening is very similar (even though it's much cleaner from the point of view of the code).

> I was still getting a decent performance

We can sync up once all the AWS services are refactored and test it out with a large amount of resources :smile:

PR #225, opened 2019-03-06
4.3.0 hotfixes

zer0x64:
PR to do small modifications required for 4.3.0 release.


j4v:
Is this because `requests` is included in boto3?

zer0x64:
> 
> 
> Is this because `requests` is included in boto3?

Requests is not used directly anymore AFAIK and I wanted to remove the "Opinel" section of the dependency!

j4v:
> Requests is not used directly anymore AFAIK

You're right, my bad :sweat_smile:

PR #229, opened 2019-03-07
Refactoring/gcp/stackdriverlogging

Remi05:
Migrated GCP's Stackdriver logging service to the new architecture. Stackdriver logging is one of the simpler services for GCP and serves as a great proof of concept for the refactoring of the architecture for this provider.

PR #230, opened 2019-03-07
Refactoring/gcp/iam

Remi05:
Migrated GCP's IAM service to the new architecture.  

Note that `gcp/resources/projects.py` and `gcp/facade/base.py` are the same as in the [Stackdriver logging PR](https://github.com/nccgroup/ScoutSuite/pull/229); they are simply shared by all services and will appear in all GCP refactoring PRs until one of the services is merged.


Remi05:
The service seems to work fine but the resource counts have changed; In the previous implementation, IAM showed a count of 8 resources where as the new implementation shows 4 resources. In both cases, there are 4 service accounts, which is shown correctly in the top navigation bar for both implementations. I thought it might be due to the keys and bindings counts but they do not seem to be set, even in the previous implementation.

Before:
iam_beforerefactoring

After:
iam_afterrefactoring

Remi05:
The build doesn't pass for Python 3.5 because of formatted strings using the `f'Text {my_variable}'` syntax which was introduced in Python 3.6. Do we still plan on supporting Python 3.5 or only 3.6 and 3.7?

j4v:
> Do we still plan on supporting Python 3.5 or only 3.6 and 3.7?

Well asyncio is available from 3.5+, but 3.6+ should be the standard 3.x on most distros. If there's a benefit to only supporting 3.6+ that's something we could consider. I'm seeing a lot of changes happening in python 3.x so if we're along for the ride (and everyone will have to be from 2020 onward) then we might as well just accept that having outdated versions of python is not acceptable anymore.

Aboisier:
> > Do we still plan on supporting Python 3.5 or only 3.6 and 3.7?
> 
> Well asyncio is available from 3.5+, but 3.6+ should be the standard 3.x on most distros. If there's a benefit to only supporting 3.6+ that's something we could consider. I'm seeing a lot of changes happening in python 3.x so if we're along for the ride (and everyone will have to be from 2020 onward) then we might as well just accept that having outdated versions of python is not acceptable anymore.

I'm not sure formatted string literals is a good enough reason to deprecate 3.5... but at the same time, it seems like there's barely anyone using 3.5. 

image

ScoutSuite/providers/gcp/configs/services.py
```diff
@@ -20,11 +22,14 @@ def __init__(self, metadata=None, thread_config=4, projects=None, **kwargs):
...
 
+        gcp_facade = GCPFacade()
+        iam_facade = IAMFacade()
+
         self.cloudresourcemanager = CloudResourceManager(thread_config=thread_config)
         self.cloudstorage = CloudStorageConfig(thread_config=thread_config)
         self.cloudsql = CloudSQLConfig(thread_config=thread_config)
         self.computeengine = ComputeEngineConfig(thread_config=thread_config)
-        self.iam = IAMConfig(thread_config=thread_config)
+        self.iam = IAM(gcp_facade, iam_facade)
```
Aboisier: Same comment as with the Stack Driver Logging PR, I'd merge the two facades.

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+from googleapiclient import discovery
+from ScoutSuite.providers.gcp.utils import MemoryCache
+
+class IAMFacade:
+    def __init__(self):
+        self._iam_client = discovery.build('iam', 'v1', cache_discovery=False, cache=MemoryCache())
```
Aboisier: [Same comment](https://github.com/nccgroup/ScoutSuite/pull/229#discussion_r263377080), I don't think this is the right place to do HTTP requests.

ScoutSuite/providers/gcp/resources/iam/service_accounts.py
```diff
@@ -0,0 +1,38 @@
...
+    ]
+
+    def __init__(self, iam_facade, project_id):
+        self.iam_facade = iam_facade
+        self.project_id = project_id
+
+    async def fetch_all(self):
+        raw_service_accounts = await self.iam_facade.get_service_accounts(self.project_id)
+        for raw_service_account in raw_service_accounts:
+            sa_id, service_account = self._parse_service_account(raw_service_account)
```
Aboisier: I'd rename `sa_id` for `service_account_id`.

ScoutSuite/providers/gcp/resources/iam/service_accounts.py
```diff
@@ -0,0 +1,38 @@
...
+            self[sa_id] = service_account
+        await self._fetch_children()
+    
+    async def _fetch_children(self):
+        for sa_id, service_account in self.items():
```
Aboisier: Same here, I'd rename `sa_id` to `service_account_id`.

ScoutSuite/providers/gcp/resources/iam/service_accounts.py
```diff
@@ -0,0 +1,38 @@
...
+            for child_name, child_class in self._children:
+                child = child_class(self.iam_facade, self.project_id, service_account['email'])
+                await child.fetch_all()
+                self[sa_id][child_name] = child
+
+    def _parse_service_account(self, raw_service_account):
```
Aboisier: [Same as here](https://github.com/nccgroup/ScoutSuite/pull/229#discussion_r263382611), would rename to `_parse`.

ScoutSuite/providers/gcp/resources/iam/bindings.py
```diff
@@ -0,0 +1,18 @@
...
+        self.project_id = project_id
+        self.service_account_email = service_account_email 
+
+    async def fetch_all(self):
+        raw_bindings = await self.iam_facade.get_bindings(self.project_id, self.service_account_email)
+        for raw_binding in raw_bindings.get('bindings', []):
+            binding_id, binding = self._parse_binding(raw_binding)
+            self[binding_id] = binding
+
+    def _parse_binding(self, raw_binding):
```
Aboisier: [Same](https://github.com/nccgroup/ScoutSuite/pull/229#discussion_r263382611), would rename to `_parse`.

ScoutSuite/providers/gcp/resources/iam/keys.py
```diff
@@ -0,0 +1,23 @@
...
+        self.project_id = project_id
+        self.service_account_email = service_account_email 
+
+    async def fetch_all(self):
+        raw_keys = await self.iam_facade.get_keys(self.project_id, self.service_account_email)
+        for raw_key in raw_keys.get('keys', []):
+            key_id, key = self._parse_key(raw_key)
+            self[key_id] = key
+
+    def _parse_key(self, raw_key):
```
Aboisier: [Same](https://github.com/nccgroup/ScoutSuite/pull/229#discussion_r263382611), would rename to `_parse`.

ScoutSuite/providers/gcp/facade/facade.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
Aboisier: [Same as here](https://github.com/nccgroup/ScoutSuite/pull/229#discussion_r263374292), would remove this line. I won't point out every occurrence but applies to all the other files.

ScoutSuite/providers/gcp/facade/facade.py
```diff
@@ -0,0 +1,18 @@
...
+
+from googleapiclient import discovery
+from ScoutSuite.providers.gcp.utils import MemoryCache
+
+class GCPFacade:
+    def __init__(self):
+        self._resourcemanager_client = discovery.build('cloudresourcemanager', 'v1', cache_discovery=False, cache=MemoryCache())
+
+    # TODO: Make truly async    
+    async def get_projects(self):
```
Aboisier: I think you could generalize this method. There is code duplication between this and [this](https://github.com/nccgroup/ScoutSuite/pull/230/files#diff-ffea74d1a65db135e82d200249fd7e5eR21).

j4v:
> it seems like there's barely anyone using 3.5

:laughing: 

There's 2 types of people, those who install dependencies on their host system and burn in hell, and those who use virtual environments. The later shouldn't care what version of 3.x is used...

Remi05:
> > > Do we still plan on supporting Python 3.5 or only 3.6 and 3.7?
> > 
> > 
> > Well asyncio is available from 3.5+, but 3.6+ should be the standard 3.x on most distros. If there's a benefit to only supporting 3.6+ that's something we could consider. I'm seeing a lot of changes happening in python 3.x so if we're along for the ride (and everyone will have to be from 2020 onward) then we might as well just accept that having outdated versions of python is not acceptable anymore.
> 
> I'm not sure formatted string literals is a good enough reason to deprecate 3.5... but at the same time, it seems like there's barely anyone using 3.5.
> 
> image

I wouldn't deprecated Python 3.5 over formatted strings, especially since you can format them differently, but if we were already going to, I would have kept this cleaner syntax 😉

ScoutSuite/providers/gcp/facade/facade.py
Remi05: Yeah, I had noticed but I wasn't sure how to do it cleanly, I think I found a decent way. Let me know what you think!

Remi05:
I replaced formatted string using the `f'Text {variable}'` with `'{} text {}'.format(var1, var2)` to remain compatible with Python 3.5!

ScoutSuite/providers/gcp/facade/facade.py
Aboisier: Yup, that's exactly what I had in mind. You'll have to make this async though (like [this commit](https://github.com/nccgroup/ScoutSuite/pull/224/commits/25967e57b29d9c66af0daf339162aa5e0492e3ea)). Michael spent a few days looking for the performance bottleneck, and I spent a day, and it was that; it makes a huge difference.

ScoutSuite/providers/gcp/facade/facade.py
```diff
@@ -0,0 +1,12 @@
+
+class Facade:
```
misg: Is this an attempt to make an abstract interface for all cloud provider facades? Seeing how really different they are, I'm not sure we can really make them homogeneous...

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,30 @@
+from googleapiclient import discovery
+from ScoutSuite.providers.gcp.utils import MemoryCache
+from ScoutSuite.providers.gcp.facade.facade import Facade
+from ScoutSuite.providers.gcp.facade.utils import GCPFacadeUtils
+
```
misg:
```suggestion
from ScoutSuite.providers.utils import run_concurrently

```

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,30 @@
...
+class IAMFacade(Facade):
+    def _build_client(self):
+        return discovery.build('iam', 'v1', cache_discovery=False, cache=MemoryCache())
+
+    # TODO: Make truly async
+    async def get_bindings(self, project_id, service_account_email):
+        resource = 'projects/{}/serviceAccounts/{}'.format(project_id, service_account_email)
+        iam_client = self._get_client()
+        response = iam_client.projects().serviceAccounts().getIamPolicy(resource=resource).execute()
```
misg:
```suggestion
        response = await run_concurrently(
                lambda: iam_client.projects().serviceAccounts().getIamPolicy(resource=resource).execute()
        )
```

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,30 @@
...
+        return response.get('bindings', [])
+
+    # TODO: Make truly async
+    async def get_keys(self, project_id, service_account_email):
+        name = 'projects/{}/serviceAccounts/{}'.format(project_id, service_account_email)
+        iam_client = self._get_client()
+        response = iam_client.projects().serviceAccounts().keys().list(name=name).execute()
```
misg:
```suggestion
        response = await run_concurrently(
                lambda: iam_client.projects().serviceAccounts().keys().list(name=name).execute()
        )
```

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,30 @@
...
+        return response.get('keys', [])
+
+    # TODO: Make truly async
```
misg: Already truly async afaik, unless there's an API call behind the assignment of `request` and/or `service_accounts_group`?

ScoutSuite/providers/gcp/resources/iam/service_accounts.py
```diff
@@ -0,0 +1,38 @@
...
+
+    async def fetch_all(self):
+        raw_service_accounts = await self.gcp_facade.iam.get_service_accounts(self.project_id)
+        for raw_service_account in raw_service_accounts:
+            service_account_id, service_account = self._parse_service_account(raw_service_account)
+            self[service_account_id] = service_account
+        await self._fetch_children()
+    
+    async def _fetch_children(self):
+        for service_account_id, service_account in self.items():
```
misg: You'll need to parallelize this loop (see https://github.com/nccgroup/ScoutSuite/pull/229/files#r263374926) and probably make a refactoring of `fetch_children` as well (would be probably worth it to define a `GCPCompositeResources` to do so, like `AWSCompositeResources` and `AzureCompositeResources` ([here](https://github.com/nccgroup/ScoutSuite/blob/28e6d1f3e4bd4b068c0b0caf31eb718743919d3c/ScoutSuite/providers/aws/resources/resources.py#L25) and [there](https://github.com/nccgroup/ScoutSuite/blob/28e6d1f3e4bd4b068c0b0caf31eb718743919d3c/ScoutSuite/providers/azure/resources/resources.py#L8))).

ScoutSuite/providers/gcp/facade/iam.py
Remi05: That was before I extracted the `get_all()` loop, I'll update the comment 😉

ScoutSuite/providers/gcp/facade/facade.py
Remi05: I'm aware, that's what the comment was for, I was waiting for Michael's PR to be merged before adding it! 😉

ScoutSuite/providers/gcp/facade/facade.py
Remi05: No it was simply for GCP, it avoids duplicating the `_get_client()` method, I can give it another name though

ScoutSuite/providers/gcp/resources/iam/service_accounts.py
Remi05: Done!

ScoutSuite/providers/gcp/facade/facade.py
Aboisier: For AWS, I called my base facade  `AWSBaseFacade`. I think it's less confusing than `Facade`.

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,30 @@
...
+        iam_client = self._get_client()
+        response = await run_concurrently(
+                lambda: iam_client.projects().serviceAccounts().keys().list(name=name).execute()
+        )
+        return response.get('keys', [])
+
+    async def get_service_accounts(self, project_id):
+        name = 'projects/{}'.format(project_id)     
+        iam_client = self._get_client()
+        request = iam_client.projects().serviceAccounts().list(name=name)
```
Aboisier: I think this should be parallelized!

```suggestion
        request = await run_concurrently(
                iam_client.projects().serviceAccounts().list(name=name)
        )
```

ScoutSuite/providers/gcp/facade/iam.py
```diff
@@ -0,0 +1,30 @@
...
+        service_accounts_group = iam_client.projects().serviceAccounts()
```
Aboisier: Should that be parallelized too?
```suggestion
        service_accounts_group = await run_concurrently(
                iam_client.projects().serviceAccounts()
        )
```

ScoutSuite/providers/gcp/facade/iam.py
Remi05: `request` and `services_accounts_group` simply build the request from my understanding. The actual fetching is done in `GCPFacadeUtils.get_all()` which is already parallelized .

ScoutSuite/providers/gcp/facade/facade.py
Remi05: Renamed it to `GCPBaseFacade`!

Remi05:
Fixed UI and added left menu for projects.

Before : 

iam_before

After : 

iam_after

ScoutSuite/providers/gcp/resources/iam/keys.py
```diff
@@ -0,0 +1,22 @@
...
+
+    async def fetch_all(self):
+        raw_keys = await self.gcp_facade.iam.get_keys(self.project_id, self.service_account_email)
+        for raw_key in raw_keys:
+            key_id, key = self._parse_key(raw_key)
+            self[key_id] = key
+
+    def _parse_key(self, raw_key):
+        key_dict = {}
+        key_dict['id'] = raw_key['name'].split('/')[-1]
```
Aboisier: Can we add a small comment here explaining where this comes from?

Remi05: Done!

PR #216, opened 2019-03-04
Refactoring/aws/cloudformation

Aboisier:
This PR migrates CloudFormation to the new architecture (see #183 ).

PR #237, opened 2019-03-08
Bugfix - CloudFormation deletion policy

vifor2:
It turns out that when calling the method get_template boto3 sometimes returns the TemplateBody as a dict, but other times as a string (because why be consistent). Also made the overall process for checking the rule more thorough.

Please also verify the [private PR](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/107).

### DIFF
--- a/ScoutSuite/providers/aws/metadata.json
+++ b/ScoutSuite/providers/aws/metadata.json
@@ -43,7 +43,7 @@
                 }
             }
         },
-            "cloudwatch": {
+        "cloudwatch": {
             "resources": {
                 "alarms": {
                     "api_call": "describe_alarms",
--- a/ScoutSuite/providers/aws/metadata.json
+++ b/ScoutSuite/providers/aws/metadata.json
@@ -43,7 +43,7 @@
                 }
             }
         },
-        "cloudwatch": {
+            "cloudwatch": {
             "resources": {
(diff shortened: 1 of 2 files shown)


vifor2:
> Do you think you could do this fix off [https://github.com/[/pull/216](https://github.com/nccgroup/ScoutSuite/pull/216)](this branch)? We're going to have naughty conflicts when we merge, and I wouldn't want to lose this fix!

Yeah sure, is there currently an issue with that branch though? When running it I'm receiving multiple errors:
`$git pull
$git reset --hard
$--provider aws --force --no-browser --thread-config 1 --service cloudformation --ruleset default.json`
image

Aboisier:
Nice catch! There was a problem with deleted stacks. It should be fixed now. Sorry about that.

vifor2:
> Nice catch! There was a problem with deleted stacks. It should be fixed now. Sorry about that.

Well... now that that branch of yours has been merged onto refactoring/resource-configs there's no more errors, but the service isn't being scanned either. Anyways, since the refactoring branch doesn't seem like it will get merged for next release y'all can still approve this to fix develop...
image

Aboisier:
Oh, that's odd. Sure, I'll just poke you when I merge this with my code to make sure we are not regressing.

PR #241, opened 2019-03-11
Feature/ec2 userdata secrets

j4v:
Improves the EC2 user data finding & implementation to check for actual secrets (not just presence of user data).

Also moves the finding to the public repo. I pushed some proprietary changes to `develop` by mistake (https://github.com/nccgroup/ScoutSuite-Proprietary/commit/37985a5358c2aaff0f72f4d1ccec7010e4297b98), but it only removes the finding / entry in the ruleset so I don't think it's an issue.


ScoutSuite/providers/aws/resources/ec2/instances.py
```diff
@@ -28,3 +30,34 @@ class EC2Instances(AWSResources):
...
+            if rsa_private_key_list:
+                secrets['Private Keys'] = rsa_private_key_list
+            word_list = []
+            for word in keywords:
+                if word in user_data.lower():
+                    word_list.append(word)
+            if word_list:
+                secrets['Flagged Words'] = word_list
+
+        return secrets
```
Aboisier: Do you think it would be a good idea to make those regexes and forbidden keywords arguments in the rule, and do the processing directly in the rule?

j4v: I could, but then I wouldn't be able to show what's being flagged within the user data or why (I think?). As user data can be a lot of content, if you just say "hey I found something, figure out what" it wouldn't be very helpful. 

The current view lists the occurrences identified, what they were identified to be (AWS credentials, RSA keys, suspicious keywords, etc.).

Aboisier: Fair enough!

ScoutSuite/output/data/html/partials/aws/services.ec2.regions.id.vpcs.id.instances.html
```diff
@@ -33,8 +33,23 @@ <h4>Network interfaces</h4>
         <div class="list-group-item" style="padding-bottom: 0 !important;">
             <h4>User data</h4>
             <div>
-                <pre><code>{{#each (split_lines user_data)}}   {{this}}<br>{{/each}}</code></pre>
+                <code>{{#each (split_lines user_data)}}   {{this}}<br>{{/each}}</code>
```
Aboisier: Does this still show up correctly in the report? I know I had a lot of trouble displaying this. I don't remember exactly why I put the `<pre>` tag, but I think it was important. hehe

j4v: The only difference `<pre></pre>` made (that I'm seeing) is remove the code color highlight.

PR #245, opened 2019-03-11
Refactoring/aws/elasticache

Aboisier:
This PR migrates ElastiCache to the new architecture, see #183 . I had to work quite a bit to do this one. Here are a few of the difficulties I encountered:
* If wasn't really working before. For example, the parameter groups and subnet groups were not showing.
* The clusters are scoped by VPC, but you can't specify a VPC Id when you call the API, and it doesn't tell you which VPC the cluster belongs to, so you have to make another call to the API.
* Because of that, the number of API calls would have exploded, so I had to implement some caching.
* I had to use locks to make sure the caching was thread safe, but the locks weren't working because our stuff is async. You have to use `asyncio.Lock` instead of `threading.Lock`, and do `async with lock: #your critical code` instead of just `with lock: #your critical code`.


j4v:
> I had to work quite a bit to do this one

How well did the new architecture handle these difficulties? Did you have to break any conventions?

> I had to implement some caching.

Would it be worth documenting this case in the wiki, in case other services might be the same?

Aboisier:
> > I had to work quite a bit to do this one
> 
> How well did the new architecture handle these difficulties? Did you have to break any conventions?

It was a bit hard to make all the pieces fit in the new architecture, but I'd say it's because the previous architecture allowed to make some hacks. If I was to implement this service from scratch, it would have been easier. I did not have to break any conventions.

> > I had to implement some caching.
> 
> Would it be worth documenting this case in the wiki, in case other services might be the same?

Good idea, I'll add a sentence or two regarding this.

Aboisier:
> I have a feeling like you've ingested a lot of caffeine in the past weeks. 

I most certainly did! 

> Nice work, good job on the comments for the more obscure parts of the code too.

Thanks!

PR #246, opened 2019-03-12
ELBv2 findings

vifor2:
PR for [this issue](https://github.com/nccgroup/ScoutSuite-Proprietary/issues/108).
Please also check out the [private PR](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/113).

Added the following findings, the third one being split into 3 findings so that the user doesn't have to investigate by himself what's wrong. I had to copy EC2's data about security groups into ELBv2 since the API call [`describe_security_groups()`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_security_groups) isn't available on the ELBv2 service.

- [Review AWS Internet Facing Load Balancers](https://www.cloudconformity.com/conformity-rules/ELBv2/internet-facing-load-balancers.html)
- [AWS ALB (ELBv2) Listener Security](https://www.cloudconformity.com/conformity-rules/ELBv2/listener-security.html)
- [AWS ELBv2 Security Groups](https://www.cloudconformity.com/conformity-rules/ELBv2/security-group.html) 
    * VPC's Default security group
    * Inbound rules do not match listeners
    * Outbound rules do not match listeners

image

P.S. I know there are quite a few layers of `for` loops, if you have a concrete idea on how to improve this please feel free to tell me so.

### DIFF
--- a/ScoutSuite/providers/aws/provider.py
+++ b/ScoutSuite/providers/aws/provider.py
@@ -106,6 +106,7 @@ class AWSProvider(BaseProvider):
                            {'AWSAccountId': self.aws_account_id})
 
     def _add_security_group_data_to_elbv2(self):
+        none = 'N/A'
         ec2_config = self.services['ec2']
         elbv2_config = self.services['elbv2']
         for region in elbv2_config['regions']:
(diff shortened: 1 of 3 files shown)


vifor2:
Forgot to mention, but the highlighting for the rules 3 to 5 is not active at the moment since the collapsibles it rely's on it broken, I have created [an issue](https://github.com/nccgroup/ScoutSuite/issues/247) to address that problem.

ScoutSuite/providers/aws/provider.py
```diff
@@ -104,6 +106,25 @@ def _add_security_group_name_to_ec2_grants(self):
                            self.add_security_group_name_to_ec2_grants_callback,
                            {'AWSAccountId': self.aws_account_id})
 
+    def _add_security_group_data_to_elbv2(self):
```
Aboisier: This method is a bit cryptic. One way of making it more readable would be to iterate through the values instead of the keys. Also, using words instead of acronyms. For example:

```py
for region in elbv2_config['regions'].values():
    for vpc in region['vpcs'].values():
        for load_balancer in vpc['lbs'].values():
            for i in range(len(load_balancer['security_groups'])):
                for security_group in vpc['security_groups'].values():
                    try:
                        if load_balancer['security_groups'][i]['GroupId'] == sg:
                             load_balancer['security_groups'][i] = vpc['security_groups'][sg]
                    except KeyError:
                        pass

                    check_security_group_rules(load_balancer, i, 'ingress')
                    check_security_group_rules(load_balancer, i, 'egress')
```

I'm not sure there's a better way of implementing this, but the five levels of nesting have a bit of a bad smell smell. Is it possible that some users might have a large amount of vpcs, load balancers and security groups? *O(n^4)*, or at least *O(n_vpcs * n_lbs * n_sgs_in_lb * n_sgs_in_vpc)*, considering the number of regions is constant.

ScoutSuite/providers/aws/services/elbv2.py
```diff
@@ -75,3 +79,16 @@ class ELBv2Config(RegionalServiceConfig):
...
+    none = 'N/A'
+    if traffic_type == 'ingress':
+        output = 'valid_inbound_rules'
+    elif traffic_type == 'egress':
+        output = 'valid_outbound_rules'
+    for protocol in lb['security_groups'][index]['rules'][traffic_type]['protocols']:
+        for port in lb['security_groups'][index]['rules'][traffic_type]['protocols'][protocol]['ports']:
+            lb['security_groups'][index][output] = True
+            if port not in lb['listeners'] and port != none:
+                lb['security_groups'][index][output] = False
```
Aboisier: Same thing here, using `.valuse()` would make the code more readable.

PR #252, opened 2019-03-13
Optimize fetching Monitor logs

j4v:
This PR optimizes fetching logs to filter for the only resource provider that's currently being used (`Microsoft.Storage`).

In the test subscription, this brings the log count from 3895 to 176, so it's a significant improvement. Nonetheless, I'm still finding that for production accounts, `resourceProvider` still includes a lot (too many) logs. Ideally we'd have to filter by `resourceURI`, which is the `resource_id` in the `activity_log` object (e.g. `/subscriptions/<subscription ID>/resourceGroups/<resource_group_name>/providers/Microsoft.Storage/storageAccounts/<storage account`). @misg do you think this is possible, as the info would have to be fetched from the Storage Accounts service...


j4v:
@misg should we close this as it is covered by https://github.com/nccgroup/ScoutSuite/pull/285?

misg:
> @misg should we close this as it is covered by #285?

Indeed, I agree!

PR #257, opened 2019-03-16
Refactoring/aws/elb

misg:
This PR aims to migrate AWS ELB service to the new architecture, as part of #183.
Please note that I got rid of the elb policies since they weren't used anywhere.


ScoutSuite/providers/aws/metadata.json
```diff
@@ -278,18 +278,12 @@
...
                     "cols": 2,
                     "path": "services.elb.regions.id.vpcs.id.elbs",
                     "callbacks": [
                         [ "match_security_groups_and_resources_callback", {"status_path": ["Scheme"], "sg_list_attribute_name": ["security_groups"], "sg_id_attribute_name": "GroupId"} ],
                         [ "get_lb_attack_surface", {} ]
                     ]
-                },
-                "elb_policies": {
-                    "cols": 2,
-                    "path": "services.elb.regions.id.elb_policies"
```
Aboisier: Doesn't this remove the elb policies page from the report?

ScoutSuite/providers/aws/resources/elb/load_balancers.py
```diff
@@ -0,0 +1,34 @@
...
+        for l in raw_load_balancer['ListenerDescriptions']:
+            listener = l['Listener']
+            load_balancer['listeners'][l['Listener']['LoadBalancerPort']] = listener
+
+        load_balancer['instances'] = []
+        for i in raw_load_balancer['Instances']:
+            load_balancer['instances'].append(i['InstanceId'])
+
+        load_balancer['attributes'] =\
+            await self.facade.elb.get_load_balancer_attributes(self.scope['region'], load_balancer['name'])
```
Aboisier: Same comment as with the SQS pr: https://github.com/nccgroup/ScoutSuite/pull/262#discussion_r266439742

I would put all the API calls in the facade and have this method do only parsing so that the separation of concerns is clear.

misg: Agreed, done. :)

ScoutSuite/providers/aws/metadata.json
misg: Yes. I grepped `elb_policies` in ScoutSuite and even though they were fetched and parsed (fetched in the old `aws/services/elb.py` file and parsed in the `parse_elb_policies_callback` method defined in `aws/provider.py`), they were not used anywhere, neither for findings nor for an attack_surface analysis. So It looked like dead code to me and  I took the decision to get rid of them. What do you think about that @j4v ?

Aboisier: That's fair!

PR #259, opened 2019-03-17
Refactoring/aws/elbv2

misg:
This PR aims to migrate AWS ELBv2 service to the new architecture, as part of #183.
Please note that I removed the `ssl_policies` resources (defined in the `metadata.json` file) since they weren't used anywhere.


ScoutSuite/providers/aws/provider.py
```diff
@@ -112,6 +111,18 @@ def _add_security_group_name_to_ec2_grants(self):
                            {'AWSAccountId': self.aws_account_id})
 
     def _add_security_group_data_to_elbv2(self):
+        def check_security_group_rules(lb, index, traffic_type):
```
Aboisier: Wait what

Aboisier: I'm so confused what is this

misg: It's just a function that Vincent defined for his findings related to ELBv2, is only used by `_add_security_group_data_to_elbv2` and was in `aws/services/elbv2.py`. So it wasn't really make sense to move it in `aws/resources/elbv2/` directory, moving it to `_add_security_group_data_to_elbv2` as a nested function where it is used makes more sense imho (at least now, waiting for a refactoring of all the post-processing logic).

Issue #226, opened 2019-03-06
Handle *very* large cloud accounts


vifor2:
Investigate and solve the following issue:
Currently the fact that a JSON file is generated for each report and then its data is loaded onto the web browser is not an issue for the majority of cases. But when Scout Suite is used agaisn't a service that is being utilized to it's full potential and has, let's say, over a million ressources, the JSON generated is way too big and therefore crashes the browser once the user tries to open the report.

Possible solution ideas:
- Move the JSON file to a local DB; not that complicated but requires a web server to fetch the content.
- Move the JSON file to the navigator's memory by using something like IndexedDB; JSON would still need to be loaded to be put in IndexedDB, a solution would be to use FileReader, but the user would have to manually select their file.

j4v:
After much thought to this I don't think the second approach will work (so I'm leaning towards the first). If you do find a way then I'm all ears.

What the first approach would probably look like:
- Instead of writing everything to a JSON, we would write it to a database
  - Most likely sqlite as it leverages a local file and is well documented, stable and easy to use
  - The data would probably be stored with a key:value format (sqlite supports this)
    - This would be relatively easy as the JSON dict is already a key-value store
    - There are libraries/wrappers (https://pypi.org/project/sqlitedict/) that make handling this easy
  - A benefit of this is that Scout wouldn't need to store all the data in memory until the end of the scan. It could write the partial data to the local db during the run, which would improve memory consumption (and probably keep latency quite similar, as it does a lot more writes then reads).
- A small web server would be needed to fetch the content from the database and serve it to the frontend
  - This is sort of what the frontend's JS already does. Currently the JSON is loaded in memory as a dictionary, and a JS function receives an index and looks up in the dictionary the value at that index.
    - The new implementation would instead make a request to the backend for that index, the backend would pull the content from the database and return it to the frontend.
      - This would likely require a bit of pagination, as let's say there's 200 buckets we'd only want to show maybe 20 at a time to the user. That would be a great improvement to the UI, as now we're just shoving everything into view
  - flask would be a good candidate for the web server, it's lightweight and can easily be run from the CLI
- The current flow is for the user to run Scout to generate an html report, and then open it in a browser. 
  - The only difference between two report html files is the name of the JS/JSON file which is hardcoded.
  - The new flow would probably be something like:
    - `Scout.py --run <params> <report name, explicit or inferred>`
      - This generates the sqlite db that stores the results
    - `Scout.py --serve <report name, explicit or inferred>`
[...]

PR #260, opened 2019-03-17
Local Database

vifor2:
Feature functional although it could use some improvements, but since there is less than 1 week before the end of the project @zer0x64 and me would like to get this merged ASAP. Also to cease the merge conflicts.

## Scroll down for TL;DR

Changes done within this PR :
- User can input `--result-format` with either `json` or `sqlite` as args.
- Imports of `scoutsuite-results/scoutsuite_results.js` and `scoutsuite-results/scoutsuite_exceptions.js` are no longer in `report.html` by default, they are instead in a partial that is added if the user chose `json` in the previous point.
- `scoutsuite.js` now checks whether an element with the id `json_format` or `sqlite_format` exists to know which functions to call.
- When choosing SQLite all of the user's data is saved as a `.db` file instead of a `.js` file, these files being more compact as well as readable by a server.
- User can input arguments related to the server (reminder: we have decided that when the user selects SQLite he must then enter a second command to run the server, it will not be launched automatically).
- Exposed all of the data saved in the `.db` file to our client through **CherryPy**.
- Modified all `.js` files to follow [standardJS](https://standardjs.com/), there are still some changes to do.
- All data except resource pages other than the first one are fetched at the start and properly viewable within the report.
- Added **Back** and **Next** buttons to change page, pressing one of those buttons disables it if you're at the page limit afterwards ([0, totalResources / pageSize - 1]), it removes the data of the previous resource from memory (run_results) and add the new page to it, it then overwrites the current template with the same one but with the new data within it.
- Page size is currently hard-coded it would be nice to give the option to the user of choosing the size but Id rather get this issue done and merged beforehand. Same with more buttons to go to first page, last page, page at a specific index, modal that tells the user to make sure his port 8000 is freed and the server is running, etc.
- I know that when we talked I said Id keep the paging buttons on both result formats but for now the buttons are disabled when viewing a report in JSON format (it should look as if nothing had happened) while the SQLite one have 2 buttons.
- Merged develop and it's new architecture into this branch (to do again it seems).

Known issues : 
- Works with **Firefox** but not with **Chrome**, will discuss about this issue tomorrow with @j4v : `jquery-3.3.1.min.js:2 Access to XMLHttpRequest at 'file:///home/vifor2/Documents/ScoutSuite/scoutsuite-report/inc-scoutsuite/sqlite.js' from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.`
- Items in the navbar are not in alphabetical order although they are in the JavaScript object (I'm bamboozled).
- Constant error when used with **Firefox** although it doesn't seem to actually cause any problem : `XML Parsing Error: syntax error
Location: file:///home/vifor2/Documents/ScoutSuite/scoutsuite-report/inc-scoutsuite/sqlite.js
Line Number 1, Column 1:`
- Initial data loading is a bit long compared to other providers because of the 16 regions per resource.

### Example with Azure
While in /ScoutSuite open two terminals, in the first one type in `azure --cli --serve` and in the second one your usual command followed by `--result-format sqlite`, you can then view the report has usually excepted for the fact that you'll have buttons to load in/out new pages.
### Example with GCP
Same as concept as Azure but replace the first command with `gcp --user-account --serve`
### Example with AWS
Same concept as previous ones but replace the first command with `aws --serve`


ScoutSuite/output/html.py
```diff
@@ -67,66 +81,47 @@ def prepare_html_report_dir(self):
 
 class Scout2Report(HTMLReport):
```
j4v: Could be a good time to remove the remaining occurrences of `scout2`

j4v:
:clap: :clap: :clap: 

> * Imports of `scoutsuite-results/scoutsuite_results.js` and `scoutsuite-results/scoutsuite_exceptions.js` are no longer in `report.html` by default, they are instead in a partial that is added if the user chose `json` in the previous point.

For the exceptions file it might make sense to keep it as a JSON file? Easier for the user to write/read/uptate that way.

> * User can input arguments related to the server (reminder: we have decided that when the user selects SQLite he must then enter a second command to run the server, it will not be launched automatically).

Yep that makes the most sense.

> will have to use pagination and only return a set of data on each requests

This change should likely be implemented outside of the whole server/db change. Currently we're outputting every object in the report. So even if we're within what the JSON can handle, if there's let's say 500 S3 buckets and the user accesses that section of the report, we'll show all of those. That's really less then ideal.

The way I see it, the `get_value_at` JavaScript function (ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js) should allow for pagination whatever the backend. That would require some changes to the left-most column in the report though. 

A couple of points:
- Considering this is very much a WIP and not functional, it would make sense not to make merges to `develop` but rather to keep it in a separate branch (`feature/226-local-database`) and make PRs against that branch?
- The branch should be keep up to date with https://github.com/nccgroup/ScoutSuite/tree/refactoring/resource-configs as that will become the mainstream architecture in a bit.
- As we plan on supporting both output formats for the foreseeable future, implementing a facade pattern for reading/writing from/to the backend would be ideal. The new architecture should make this a bit simpler then the current one.

ScoutSuite/output/html.py
zer0x64: No problem, will do it!

zer0x64:
>As we plan on supporting both output formats for the foreseeable future, implementing a facade pattern for reading/writing from/to the backend would be ideal.
I believe this is already done for the writing part somehow, as split JavaScriptReaderWriter into a parent and a child class, so now the parent and its exposed funtions is independent of the underlying data format. Altough It might be a better idea to choose which child class to use in a constructor function(like a factory)?

j4v:
> Altough It might be a better idea to choose which child class to use in a constructor function(like a factory)

That would also work nicely :+1:

zer0x64:
Just reaised the issue hasn't been linked yet: https://github.com/nccgroup/ScoutSuite/issues/226

zer0x64:
Server is "done", but might need adjusting if it doesn't go well with the report. So, what's left:

- [x] Server
- [ ] Report
- [ ] Save data during the run(not sure it's a great idea until SQLiteDict can support nested dicts)
- [ ] Ask for IP/port of the report if default is not used

vifor2:
Here to give a sign of life to anyone wondering what's going on, first post of this PR has been updated a bit btw.

So, I may have underestimated the architecture of the application on the `javascript` side, although there are some great things being done, there is also some improvements to be made. Rather than tackling some refactoring I'm trying to implement this feature as is, although it is not looking pretty (I did change all the lines so that they respect standardJS though since my IDE was killing me). I'll try to clean up the ugly things I'm doing atm afterwards.

TL;DR I'm moving forward slowly, it's been taking me quite some time to achieve things that seemed very simple at first glance and I'm sorry for that.

Here's the current state of things, the data on this report is being read from the **CherryPy** server :

vifor2:
I think I'll switch my way of doing things around, instead of trying to fetch from the server only the info needed in each function (and sometimes fetching the same info multiple times) I'll do one first big fetch to have an object that resembles the original `run_results` with the exception of the objects that should be scaling with bigger environments, i.e. objects under `run_results.services.[service]` that are not filters or findings. Thinking of it, that's really what I should've done from the get go.

vifor2:
Yet another update. The whole feature of changing pages & stuff works, but the initial loading of data feels too long when using AWS vs Azure. To solve this I propose we do more work on the server side to send a big dict (send everything in run_results except for run_results.[SERVICE].key where key is not filters, findings, doesn't end by '_count') to the client, this way the client will only do one request at the start instead of x*y*z*v*w requests. 

Also if I could have an URL where if I do a GET I receive the whole sub dict it would be really useful and I could get rid of my monstrous code chunk that tries to fetch something, check if it has keys or not and maybe or not goes deeper to do the same kind of requests again and again.

Do you think you could take care of those modifications server-side @zer0x64 ? 
TL;DR : 
- data?key=run_results or name of your choosing returns everything except resources
- data?dict returns the dict as is, not just one layer, idc about type and keys attributes, I just want the raw stuff!

Paging in action, yes some data is missing from the page.

zer0x64:
Here's the draft page in the wiki to document what is done, what is not, and ideas on there to go from now:
https://github.com/nccgroup/ScoutSuite/wiki/(Draft)-Server

Scout.py
```diff
@@ -1,39 +1,8 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
Aboisier: We started removing those since it's implicit with the most recent versions of python.
``suggestion
```

File: ScoutSuite/__init__.py [hunk 3]
```diff
@@ -1,15 +1,9 @@
 # -*- coding: utf-8 -*-
 
 __author__ = 'NCC Group'
-__version__ = '5.0.0-beta'
+__version__ = '5.0.0'
```
Aboisier: Not sure this PR promotes the version to a stable version 😆 

```suggestion
__version__ = '5.0.0-beta'
```

ScoutSuite/__main__.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/__main__.py
```diff
@@ -0,0 +1,210 @@
...
+    # Open the report by default
+    if not args.get('no_browser'):
+        print_info('Opening the HTML report')
+        url = 'file://%s' % os.path.abspath(html_report_path)
+        webbrowser.open(url, new=2)
+
+    return 0
+
+
+def generate_report_name(provider_code, args):
```
Aboisier: Can we move this to a util file? I feel like it doesn't belong to the `__main__.py`.

ScoutSuite/__main__.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # Save config and create HTML report
+    html_report_path = report.save(cloud_provider, exceptions, args.get('force_write'),
+                                   args.get('debug'))
+
+    # Open the report by default
+    if not args.get('no_browser'):
+        print_info('Opening the HTML report')
+        url = 'file://%s' % os.path.abspath(html_report_path)
+        webbrowser.open(url, new=2)
```
Aboisier: Seems like this does not work if the server is not running. Couldn't we start the server automatically at this point, or add an argument that does that? Having to use two terminals and two commands is a bit of a pain.

ScoutSuite/__main__.py
j4v: This has already been removed in `develop`, merging this branch will be **fun**.

ScoutSuite/core/cli_parser.py
```diff
@@ -239,9 +235,32 @@ def _init_common_args_parser(self):
...
+                                 "This must be used when the results are exported as an sqlite database.")
+        parser.add_argument('--host',
+                            dest="host_ip",
+                            default="127.0.0.1",
+                            help="Address on which you want the server to listen. Defaults to localhost.")
+        parser.add_argument('--port',
+                            dest="host_port",
+                            type=int,
+                            default=8000,
+                            help="Port on which you want the server to listen. Defaults to 8000.")
```
Aboisier: Can we put the server related stuff in a sub-parser or something like that?

ScoutSuite/core/conditions.py
```diff
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/core/conditions.py
```diff
@@ -1,3 +1,5 @@
...
+
```
Aboisier:
```suggestion
```

ScoutSuite/core/ruleset.py
```diff
@@ -1,8 +1,10 @@
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/core/ruleset.py
```diff
@@ -1,8 +1,10 @@
...
+
```
Aboisier:
```suggestion
```

ScoutSuite/core/ruleset.py
```diff
@@ -45,27 +47,27 @@ def __init__(self,
...
+        params = {'aws_account_id': aws_account_id}
         if ruleset_generator:
             self.prepare_rules(attributes=['description', 'key', 'rationale'], params=params)
         else:
             self.prepare_rules(ip_ranges=ip_ranges, params=params)
 
     def load(self, rule_type, quiet=False):
         """
-        Open a JSON file defining a ruleset and load it into a Ruleset object
+        Open a JSON file definiting a ruleset and load it into a Ruleset object
```
Aboisier:
```suggestion
        Open a JSON file defining a ruleset and load it into a Ruleset object
```

ScoutSuite/core/ruleset.py
```diff
@@ -203,7 +205,7 @@ def find_file(self, filename, filetype='rulesets'):
 
 class TmpRuleset(Ruleset):
 
-    def __init__(self, cloud_provider, rule_dirs=None, rule_filename=None, rule_args=None, rule_level='danger'):
+    def __init__(self, cloud_provider='aws', rule_dirs=None, rule_filename=None, rule_args=None, rule_level='danger'):
```
Aboisier: Why defaulting to aws?

ScoutSuite/core/server.py
```diff
@@ -0,0 +1,107 @@
...
+
+count_re = re.compile(r".*_count$")
+
+class Server(object):
+    def __init__(self, filename):
+        self.results = SqliteDict(filename)
+
+    @cherrypy.expose()
+    @cherrypy.tools.json_out()
+    def summary(self):
```
Aboisier: This is a bit cryptic, can we add a docstring? Unit tests would also be appropriate.

ScoutSuite/core/server.py
```diff
@@ -0,0 +1,107 @@
+from sqlitedict import SqliteDict
+import cherrypy
+import cherrypy_cors
+
+import re
+
+count_re = re.compile(r".*_count$")
+
+class Server(object):
```
Aboisier: This class lacks docstrings, and the method names are also very generic. It's a bit hard to understand what is going on. I think unit tests would be very appropriate and easy to implement.

ScoutSuite/core/server.py
```diff
@@ -0,0 +1,107 @@
...
+                'server.socket_port': port,
+        })
+        cherrypy.quickstart(Server(database_filename), "/api", config=config)
+
+    @staticmethod
+    def get_item(data, key):
+        if not key:
+            return data
+
+        keyparts = key.split('¤')
```
Aboisier: Are we actually using that or are we still looking for a solution haha? One would be to escape the keys and then use dots.

ScoutSuite/output/console.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/output/console.py
```diff
@@ -0,0 +1,135 @@
...
+
```
Aboisier:
```suggestion
```

ScoutSuite/output/data/html/partials/aws/left_menu_for_region.html
```diff
@@ -1,9 +1,9 @@
 <!-- Generic left-menu link partial for region-then-resource -->
 <script id="left_menu_for_region.partial" type="text/x-handlebars-template">
-  <div id="services.cloudtrail.regions.id.trails.list">
+  <div id="services.cloudtrail.regions.id.trails.list" />
   <div class="list-group" >
     <div class="list-group-item">
-      <a href="javascript:showRowWithItems('services.{{service_name}}.regions.id.{{resource_type}}')">Show all <span class="badge float-right btn-info">{{get_value_at 'metadata' service_group service_name 'resources' resource_type 'count'}}</span></a>
+      <a href="javascript:showRowWithItems('services.{{service_name}}.regions.id.{{resource_type}}')">Show all <span class="badge float-right">XXX</span></a>
```
Aboisier: Why did we lose the count?

ScoutSuite/output/data/html/partials/aws/left_menu_for_region.html
```diff
@@ -1,9 +1,9 @@
 <!-- Generic left-menu link partial for region-then-resource -->
 <script id="left_menu_for_region.partial" type="text/x-handlebars-template">
-  <div id="services.cloudtrail.regions.id.trails.list">
+  <div id="services.cloudtrail.regions.id.trails.list" />
```
Aboisier: Why did this slash appear?

ScoutSuite/output/data/html/partials/aws/services.ec2.regions.vpcs.security_groups.resource_list.html
```diff
@@ -3,7 +3,7 @@
       {{#if resources}}
         <div class="accordion">
           <li>
-            <h5 class="list-group-item-heading accordion-heading">{{ make_title service }} {{ make_title resource_type }}
+            <h5 class="list-group-item-heading accordion-heading">{{ makeTitle service }} {{ makeTitle resource_type }}
                 <!--{{> count_badge count=resources.length}}-->
```
Aboisier:
```suggestion
```

ScoutSuite/output/data/html/partials/aws/services.ec2.regions.vpcs.security_groups.resource_list.html
```diff
@@ -14,7 +14,7 @@ <h5 class="list-group-item-heading accordion-heading">{{ make_title service }} {
                   <li class="list-group-item-text">
                     <a href="javascript:showObject('services.{{../service}}.regions.{{../region}}.vpcs.{{../vpc}}.{{../resource_type}}.{{this}}')">
                         {{this}}
-                        <!--{{get_value_at 'services' ../service 'regions' ../region 'vpcs' ../vpc ../resource_type this 'name'}}-->
+                        <!--{{getValueAt 'services' ../service 'regions' ../region 'vpcs' ../vpc ../resource_type this 'name'}}-->
```
Aboisier:
```suggestion
```

ScoutSuite/output/data/html/partials/aws/services.rds.regions.id.vpcs.id.instances.html
```diff
@@ -9,8 +9,7 @@ <h4>Information</h4>
           <ul>
             <li class="list-group-item-text">Region: {{region}}</li>
             <li class="list-group-item-text">Engine: {{Engine}}</li>
-            <li class="list-group-item-text">Status: {{make_title DBInstanceStatus}}</li>
-            <li class="list-group-item-text">Is read replica: {{is_read_replica}}</li>
+            <li class="list-group-item-text">Status: {{makeTitle DBInstanceStatus}}</li>
```
Aboisier: I think the `is_read_replica` line was lost in a merge somewhere. 
```suggestion
            <li class="list-group-item-text">Is read replica: {{is_read_replica}}</li>
            <li class="list-group-item-text">Status: {{makeTitle DBInstanceStatus}}</li>
```

ScoutSuite/output/data/html/partials/gcp/services.cloudresourcemanager.bindings.html
```diff
@@ -1,12 +1,12 @@
 
 <!-- Cloud Resources Manager bindings' partial -->
-<script id="services.cloudresourcemanager.projects.id.bindings.partial" type="text/x-handlebars-template">
+<script id="services.cloudresourcemanager.bindings.partial" type="text/x-handlebars-template">
```
Aboisier:
```suggestion
<script id="services.cloudresourcemanager.projects.id.bindings.partial" type="text/x-handlebars-template">
```

ScoutSuite/output/data/html/partials/gcp/services.cloudresourcemanager.bindings.html
```diff
@@ -1,12 +1,12 @@
```
Aboisier: I think there was a bad merge here. The `.projects.` was added in one of Remi's refactoring.

ScoutSuite/output/data/html/partials/gcp/services.cloudsql.instances.html
```diff
@@ -0,0 +1,40 @@
+
```
Aboisier: It seems like there was a bad merge here too. The `ScoutSuite/output/data/html/partials/gcp/services.cloudsql.projects.id.instances.html` file was deleted, but I it's the most recent one.

ScoutSuite/output/data/html/partials/gcp/services.iam.service_accounts.html
```diff
@@ -1,26 +1,26 @@
```
Aboisier: Same here, I think we need to revive the `.projects.id.`

ScoutSuite/output/data/html/partials/gcp/services.stackdriverlogging.sinks.html
```diff
@@ -1,24 +1,24 @@
```
Aboisier: I suspect a bad merge here too

ScoutSuite/output/data/html/report.html
```diff
@@ -9,7 +9,7 @@
 
     <title>Scout Suite Report</title>
 
-    <!--NCC favicon as base64-->
+    <!-- NCC favicon -->
```
Aboisier: 💯

ScoutSuite/output/data/html/report.html
```diff
@@ -152,23 +152,30 @@ <h4>
...
-            load_metadata();
+            loadMetadata();
         </script>
 
+        <!-- Paging section -->
+        <div id="section_paging-div" style="display:flex;flex-direction:column">
+            <h2 id="section_paging-h2"></h2>
+            <div id="paging_buttons" style="margin:auto">
+                <button id="page_backward" type="button" class="btn btn-secondary" disabled><i class="fa fa-arrow-circle-left"></i> Back</button>
+                <!--button type="button" class="btn btn-secondary"><i class="fa fa fa-times-circle"></i> X</button-->
```
Aboisier: Should we uncomment or remove this?

ScoutSuite/output/data/html/report.html
```diff
@@ -152,23 +152,30 @@ <h4>
...
+                <button id="page_forward" type="button" class="btn btn-secondary">Next <i class="fa fa-arrow-circle-right"></i></button>                
+            </div>
+        </div>
```
Aboisier: Can we move the inline css to a stylesheet?

ScoutSuite/output/data/inc-scoutsuite/pagination.js
```diff
@@ -0,0 +1,162 @@
+var defaultPageSize = 2
+var regionPageSize = 1
+var reCount = new RegExp('_count+$')
```
Aboisier: Why is there a `+`? This will match `_countttttttttt`. It probably won't cause any problem, but it seems like this is not necessary.

```suggestion
var reCount = new RegExp('_count$')
```

ScoutSuite/output/data/inc-scoutsuite/pagination.js
```diff
@@ -0,0 +1,162 @@
+var defaultPageSize = 2
+var regionPageSize = 1
```
Aboisier: Can we use  `const`?
```suggestion
const regionPageSize = 1
```

ScoutSuite/output/data/inc-scoutsuite/pagination.js
```diff
@@ -0,0 +1,162 @@
+var defaultPageSize = 2
```
Aboisier: Same comment as above.
```suggestion
const defaultPageSize = 2
```

ScoutSuite/__main__.py
vifor2: Ugh yeah, I'm trying to merge develop into this in a separate branch to not break the testing of the functionality here. But the fact that : I changed all the lines in .js files and changed all the `scout2` s to `scoutsuite` while @j4v changed them to `scout`, on develop `aws_account_id` was changed to `account_id` which is a very small deal, but since I changed everything in scoutsuite.js it is a slighty larger deal as the IDE doesn't highlight the merge conflicts properly anymore 😂,  the stuff moved from `__main__` to `scout.py` on develop while new stuff was added on our side in __main__, just to name a few....

ScoutSuite/output/data/html/report.html
vifor2: Ugh this is here because of I copied the CSV download buttons which already had inline CSS, what about we remove the inline CSS here when "we" will remove all the inline CSS? 😃

ScoutSuite/output/data/inc-scoutsuite/pagination.js
```diff
@@ -0,0 +1,162 @@
...
+  let pageSize = pageInfo[0]
+  let pageIndex = pageInfo[1]
+  pageIndex += indexDiff
+  // getResourcePageSqlite being called in both statements is intentional, I want events to happen in this order to
+  // prevent the user from cliking on next page multiple times and going out of bounds and I want to call loadConfig
+  // to regenerate the page after Iv'e loaded SQLite data
+  if (indexDiff === 0) {
+    if (pathArray.length === 3) {
+      getResourcePageSqlite(pageIndex, pageSize, pathArray[1], pathArray[2])
+    } else if (pathArray.length === 5) {
```
Aboisier: I'm a bit confused by the pagination logic. It seems like there is AWS-specific applied to all the providers (I'm referring to the regions). I'm a bit confused by the `pathArray.length === 3` and `pathArray.length === 5`. I'm not sure where the numbers come from, and I feel like it could easily break when introducing new providers.

ScoutSuite/output/data/inc-scoutsuite/pagination.js
vifor2: You're right, @zer0x64 told me to, I'll remove it (I'm a noob at regex)

ScoutSuite/output/data/html/report.html
vifor2: Yes will remove!

ScoutSuite/output/data/html/partials/aws/services.rds.regions.id.vpcs.id.instances.html
vifor2: Not merged yet here but it is on our broken merging branch!

ScoutSuite/output/data/inc-scoutsuite/pagination.js
Aboisier: Ok, I noticed the think which checks if the path has regions. Is there a way to make the pagination system more generic so we don't have to have provider-specific logic?

ScoutSuite/output/data/inc-scoutsuite/provider.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Get the whole config dictionary
- * @returns {{account_id, last_run, metadata, provider_code, provider_name, service_groups, service_list, services, sg_map, subnet_map}|*}
+ * @returns {{aws_account_id, last_run, metadata, provider_code, provider_name, service_groups, service_list, services, sg_map, subnet_map}|*}
```
Aboisier: :o no

ScoutSuite/output/data/inc-scoutsuite/pagination.js
vifor2: When there are no regions the path looks like `services.[service].[resource]` (length of 3) while when there are regions it looks like  `services.[service].regions.[region].[resource]` (length of 5). Atm we only have AWS that uses regions, but not all services of AWS use regions.

I'll change it to verify whether the third element is `regions` or not then.

ScoutSuite/output/report_file.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/output/report_file.py
```diff
@@ -0,0 +1,11 @@
...
+
```
Aboisier:
```suggestion
```

ScoutSuite/output/result_encoder.py
```diff
@@ -0,0 +1,146 @@
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/output/report_file.py
vifor2: ok

ScoutSuite/output/result_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    JSON encoder class
+    """
+
+    def default(self, o):
+        try:
+            if type(o) == datetime.datetime:
+                return str(o)
+            else:
+                # remove unwanted attributes from the provider object during conversion to json
```
Aboisier: I think we'll be able to delete that when #183 will be fully merged, since those won't exist anymore.

ScoutSuite/providers/__init__.py
```diff
@@ -1,8 +1,10 @@
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/providers/__init__.py
```diff
@@ -1,8 +1,10 @@
...
+
```
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/authentication_strategy.py
```diff
@@ -2,28 +2,16 @@
...
-            sts_client = session.client('sts')
-            sts_client.get_caller_identity()
-
-            return credentials
+        if credentials.get('access_key') is None:
+            raise AuthenticationException()
 
-        except Exception as e:
-            raise AuthenticationException(e)
+        return credentials
```
Aboisier: Why was that changed :/?

ScoutSuite/providers/aws/configs/services.py
```diff
@@ -1,39 +1,36 @@
...
-from ScoutSuite.providers.aws.resources.iam.base import IAM
-from ScoutSuite.providers.aws.resources.rds.base import RDS
-from ScoutSuite.providers.aws.resources.redshift.base import Redshift
-from ScoutSuite.providers.aws.resources.route53.base import Route53
-from ScoutSuite.providers.aws.resources.s3.base import S3
-from ScoutSuite.providers.aws.resources.ses.base import SES
-from ScoutSuite.providers.aws.resources.sns.base import SNS
-from ScoutSuite.providers.aws.resources.sqs.base import SQS
-from ScoutSuite.providers.aws.resources.vpc.base import VPC
+# -*- coding: utf-8 -*-
```
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/configs/services.py
```diff
@@ -1,39 +1,36 @@
...
+
```
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/facade/ec2.py
```diff
@@ -55,17 +55,7 @@ class EC2Facade(AWSBaseFacade):
             'ec2', region, self.session, 'describe_network_interfaces', 'NetworkInterfaces', Filters=filters)
 
     async def get_volumes(self, region: str):
```
Aboisier: I think there was a bad merge here

ScoutSuite/providers/aws/provider.py
```diff
@@ -16,36 +19,30 @@ class AWSProvider(BaseProvider):
...
 
-        self.metadata_path = '%s/metadata.json' % os.path.split(os.path.abspath(__file__))[0]
+        self.metadata_path = '%s/metadata.json' % os.path.split(
+            os.path.abspath(__file__))[0]
 
         self.sg_map = {}
         self.subnet_map = {}
 
         self.profile = profile
+        self.aws_account_id = None
```
Aboisier: That was renamed to `account_id`... merge problem?

ScoutSuite/providers/aws/provider.py
```diff
@@ -114,10 +111,10 @@ def _add_security_group_name_to_ec2_grants(self):
         """
         self._go_to_and_do(self.services['ec2'],
                            ['regions', 'vpcs', 'security_groups', 'rules',
-                            'protocols', 'ports', 'security_groups'],
+                               'protocols', 'ports', 'security_groups'],
                            [],
                            self.add_security_group_name_to_ec2_grants_callback,
-                           {'AWSAccountId': self.account_id})
+                           {'AWSAccountId': self.aws_account_id})
```
Aboisier: Same here, should be `account_id`

ScoutSuite/providers/aws/provider.py
```diff
@@ -183,14 +180,14 @@ def add_security_group_name_to_ec2_grants_callback(self, current_config, path, c
...
-                        current_path.index('security_groups') + 1)]
+                    current_path.index('security_groups') + 1)]
                 target.append(sg_id)
             ec2_grant['GroupName'] = get_value_at(
                 self.services['ec2'], target, 'name')
 
     @staticmethod
     def _process_cloudtrail_trails(cloudtrail_config):
-        print_info('Processing CloudTrail config')
+        print_info('Processing CloudTrail config...')
```
Aboisier: The `...` were removed.
```suggestion
        print_info('Processing CloudTrail config')
```

ScoutSuite/core/cli_parser.py
vifor2: @zer0x64 your job (if we do that)

ScoutSuite/core/server.py
vifor2: @zer0x64

ScoutSuite/output/data/html/partials/aws/left_menu_for_region.html
vifor2: I think it got removed but since this PR is in the past GitHub tells us this was removed in the future

ScoutSuite/providers/aws/authentication_strategy.py
vifor2: it wasn't!!!!!!!

ScoutSuite/__main__.py
vifor2: Oh right 2 terminals, sorry I wrote that because it's easier when developing, but the user can use the same terminate to generate the report and then --serve and open the report to view it. When talking with @j4v we chose to keep this in 2 steps if irc.

ScoutSuite/core/cli_parser.py
zer0x64: This is kinda of a touchy subject, because
1. Theorically, the subparsers are meant to be providers
2. The way it is done right now, because of the execution order, you need to log in before being able to serve the report. This would require some heavier changes for now than what it's actually worth, because it touches how to provider object is created

PR #262, opened 2019-03-18
Refactoring/aws/sqs

misg:
This PR aims to migrate AWS SQS service to the new architecture, as part of #183.


ScoutSuite/providers/aws/resources/sqs/service.py
```diff
@@ -0,0 +1,39 @@
...
+        # TODO: parallelize this async loop:
+        for queue_url in queue_urls:
+            id, queue = await self._parse_queue(queue_url)
+            self[id] = queue
+
+    async def _parse_queue(self, queue_url):
+        queue = {'QueueUrl': queue_url}
+        attributes = await self.facade.sqs.get_queue_attributes(
+            self.scope['region'], queue_url, ['CreatedTimestamp', 'Policy', 'QueueArn', 'KmsMasterKeyId']
+        )
```
Aboisier: In my other PRs, I usually tried to put all the API calls in the facade call. This way, all the data is retrieved at the same place (in the facade), and the parse method really only does parsing. I think it's a good convention to follow, it respects the separation of concerns principle.

misg: You're perfectly right (and to be honest, I wasn't really satisfied with that)! I just fixed the separation of concerns. :)

misg:
> Looks good! I didn't approve right away because think it's really important to separate the concerns. Also the build is broken what's up with that.

image
:sob:

PR #264, opened 2019-03-18
Refactoring/aws/iam

Aboisier:
This PR migrates the IAM service to the new architecture, see #183 .

EDIT:
The resources count changed a bit because inline policies were included in the policies count. This did not make sense because, in the dropdown, you would see, _Policies (17)_, (15 managed policies and 2 inline, for example), but the policies page would only show the managed policies.


ScoutSuite/providers/aws/facade/iam.py
```diff
@@ -0,0 +1,203 @@
...
+                    'name', profile['InstanceProfileName'])
+
+            # Get trust relationship
+            role['assume_role_policy'] = {}
+            role['assume_role_policy']['PolicyDocument'] = role.pop(
+                'AssumeRolePolicyDocument')
+
+        return roles
+
+    async def get_password_policy(self):
```
vifor2: It was likely already like this before you moved it, but this method doesn't have a `_` in front of it while other getters have one it seems :sweat_smile:

ScoutSuite/providers/aws/provider.py
```diff
@@ -78,7 +78,7 @@ def preprocessing(self, ip_ranges=None, ip_ranges_name_key=None):
 
         self._add_cidr_display_name(ip_ranges, ip_ranges_name_key)
         self._merge_route53_and_route53domains()
-        self._match_iam_policies_and_buckets()
+        # self._match_iam_policies_and_buckets() # TODO: Uncomment this!
```
vifor2: :thinking_face: j/k this can be done afterwards :slightly_smiling_face:

ScoutSuite/providers/aws/facade/utils.py
```diff
@@ -8,25 +8,34 @@ class AWSFacadeUtils:
...
         for page in paginator:
-            resources.extend(page[key])
+            for entity in entities:
+                resources[entity].extend(page[entity])
 
         return resources
 
     @staticmethod
-    def get_client(service: str, region: str, session: boto3.session.Session):
+    def get_client(service: str, session: boto3.session.Session, region: str = None):
```
vifor2: Only leaving picky insignificant comment because I want you to know someone read your ~50 commits PR :heart: There shouldn't be a space in front or after the equals sign I believe : `str=None`

ScoutSuite/providers/aws/facade/iam.py
Aboisier: The `_` is a convention (defined in PEP 8) which means _"you shouldn't use this outside of this class"_, or _"for internal use only"_, by opposition. This getter is used by `PasswordPolicy`'s `fetch_all`, hence the non-prefixed name.

ScoutSuite/providers/aws/provider.py
Aboisier: Woopsi, good think I put a todo hehe! Thanks!

ScoutSuite/providers/aws/facade/utils.py
Aboisier: WOW thanks, means a lot 💓 . 
```suggestion
    def get_client(service: str, session: boto3.session.Session, region: str=None):
```

Aboisier:
> 💯 💯 💯
Thanks for the review!

> EDIT: Except for the part where the tests now fail image May need to merge @misg's refactor of ELBv2 into this branch to solve the issue : `ImportError: No module named 'ScoutSuite.providers.aws.services.elbv2'`

It was just a bad merge actually. :'(

ScoutSuite/providers/aws/facade/utils.py
```diff
@@ -38,5 +47,5 @@ def get_client(service: str, region: str, session: boto3.session.Session):
         """
 
         # TODO: investigate the use of a mutex to avoid useless creation of a same type of client among threads:
-        client = session.client(service, region_name=region)
+        client = session.client(service, region_name=region) if region is not None else session.client(service)
```
misg:
```suggestion
        client = session.client(service, region_name=region) if region else session.client(service)
```

ScoutSuite/providers/aws/resources/iam/service.py
```diff
@@ -0,0 +1,134 @@
...
+
+    async def fetch_all(self, credentials, regions=None, partition_name='aws'):
+        # TODO: This should not be set here, the facade should be injected and already authenticated
+        self.facade._set_session(credentials)
+        await self._fetch_children(self, {})
+
+        # We do not want the report to count the password policies as resources, they aren't really resources.
+        self['password_policy_count'] = 0
+
+    async def finalize(self):
```
misg: I think this function could be non-async?

ScoutSuite/providers/aws/resources/iam/users.py
```diff
@@ -0,0 +1,20 @@
+from ScoutSuite.providers.aws.resources.resources import AWSResources
+
+
+class Users(AWSResources):
+    async def fetch_all(self, **kwargs):
+        raw_users = await self.facade.iam.get_users()
+        for raw_user in raw_users:
+            name, resource = self._parse_user(raw_user)
+              
+            if name in self:
```
misg: Can we really end up with duplicates in `raw_users` ?!

ScoutSuite/providers/aws/resources/iam/service.py
Aboisier: It could indeed, but it is awaited (in the base provider I think), so if I remove the async, we'll get a _"can't await None"_ or some error like that.

ScoutSuite/providers/aws/resources/iam/users.py
Aboisier: I'm not sure... it was like that before, so I left it. :/ Do you think we should remove this?

ScoutSuite/providers/aws/facade/iam.py
```diff
@@ -0,0 +1,203 @@
...
+        client = AWSFacadeUtils.get_client('iam', self.session)
+        response = client.generate_credential_report()
+
+        if response['State'] != 'COMPLETE':
+            print_error('Failed to generate a credential report.')
+            return []
+
+        report = client.get_credential_report()['Content']
+
+        # The report is a CSV string. The first row contains the name of each column. The next rows
```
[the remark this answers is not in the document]
Aboisier: Yea, I hesitated with this one, but since AWS's output is very sketchy (aka a CSV, for real?), I decided to at least parse the CSV. This way, the facade returns a nice object the caller can work with.

ScoutSuite/providers/aws/resources/iam/users.py
misg: Nevermind, let's not remove it to avoid unexpected side effects...

PR #270, opened 2019-03-20
Refactoring/aws/s3

Aboisier:
This PR migrates S3 to the new architecture, see #183 .

What changed (apart from like changing the architecture):
- I removed a property named `logging_stuff` which contained what I identified as both redundant and irrelevant data.
- About 100 lines of dead code were removed


ScoutSuite/providers/aws/facade/s3.py
```diff
@@ -0,0 +1,185 @@
...
+from ScoutSuite.providers.utils import run_concurrently
+from ScoutSuite.providers.utils import get_non_provider_id
+from ScoutSuite.core.console import print_error, print_exception
+
+
+class S3Facade(AWSBaseFacade):
+
+    async def get_buckets(self):
+        client = AWSFacadeUtils.get_client('s3', None, self.session)
+        buckets = (await run_concurrently(lambda: client.list_buckets()))['Buckets']
```
misg:
```suggestion
        buckets = (await run_concurrently(client.list_buckets))['Buckets']
```

ScoutSuite/providers/aws/facade/s3.py
```diff
@@ -0,0 +1,185 @@
...
+
+        for bucket in buckets:
```
misg: I think this loop should be parallelized, Xavier told us about an AWS account with millions of buckets.. Parallelizing the fetching of all these attributes should be necessary.

ScoutSuite/providers/aws/resources/s3/service.py
```diff
@@ -0,0 +1,52 @@
+from ScoutSuite.providers.aws.facade.facade import AWSFacade
+from ScoutSuite.providers.aws.resources.regions import Regions
+from ScoutSuite.providers.aws.resources.resources import AWSResources, AWSCompositeResources
+
+from ScoutSuite.providers.utils import get_non_provider_id
+
+
+class Bucket(AWSResources):
```
misg:
```suggestion
class Buckets(AWSResources):
```

ScoutSuite/providers/aws/resources/s3/service.py
```diff
@@ -0,0 +1,52 @@
...
+        raw_bucket['CreationDate'] = str(raw_bucket['CreationDate'])
+
+        # # If requested, get key properties
+        raw_bucket['id'] = get_non_provider_id(raw_bucket['name'])
+        return raw_bucket['id'], raw_bucket
+
+
+class S3(AWSCompositeResources):
+    _children = [
+        (Bucket, 'buckets')
```
misg:
```suggestion
        (Buckets, 'buckets')
```

ScoutSuite/providers/aws/facade/s3.py
misg: (I can take care of this one, if you want.)

ScoutSuite/providers/aws/resources/s3/service.py
Aboisier: Dammit. This convention makes so much sense, but it's so easy to forget.

ScoutSuite/providers/aws/facade/s3.py
Aboisier: I think I would leave a todo and come back to later. We'll have to do some batching because it wouldn't be smart to start 1 million requests simultaneously. I'm going to try and create a generic equivalent of `run_concurrently` for task lists.

j4v: @Aboisier sounds good. If it's something you'll leave to the side please open an issue to track it. There's enough forgotten `TODO`s in the code as is!

Aboisier: We've been leaving a lot of Todos, we plan to address them all before we merge the refactoring branch, don't worry!

Issue #14, opened 2018-07-03
"Backup disabled" shouldn't be flagged for read-only replicas for RDS


imsky:
If a read replica's backup period is modified, the following message can appear: "DB Backups not supported on a read replica for engine postgres"

PR #271, opened 2019-03-21
Refactoring/aws/rds

Aboisier:
This PR migrates RDS to the new architecture, see #183 .

Here are some side effects of the migration:
* The data structure now contains _vpcs_ even if there are no resources.


j4v:
@Aboisier as part of this PR could you please confirm/resolve the following open issue https://github.com/nccgroup/ScoutSuite/issues/14?

PR #274, opened 2019-03-21
Refactoring/aws/dynamodb

Aboisier:
This PR migrates DynamoDB to the new architecture, see #183 . This is part of a proprietary feature, please also review the [proprietary PR](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/116).

**Interesting stuff:**
~The get for one table requires 3 API calls (getting the table info, the manual backups info and the continuous backups). I batched the requests instead of just parallelizing all the requests. This will mitigate the risk of being throttled.~
EDIT: The batching was actually reverted. We realized the number of concurrent requests was already limited by the number of threads, so this became superfluous.

### DIFF
--- a/ScoutSuite/providers/aws/configs/services.py
+++ b/ScoutSuite/providers/aws/configs/services.py
@@ -24,11 +24,11 @@ from ScoutSuite.providers.base.configs.services import BaseServicesConfig
 
 try:
     from ScoutSuite.providers.aws.services.config_private import ConfigConfig
-    from ScoutSuite.providers.aws.services.dynamodb_private import DynamoDBConfig
+    from ScoutSuite.providers.aws.resources.dynamodb.service_private import DynamoDB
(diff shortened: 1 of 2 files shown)


ScoutSuite/providers/aws/configs/services.py
```diff
@@ -23,12 +23,12 @@
...
 
 try:
+    from ScoutSuite.providers.aws.resources.dynamodb.service_private import DynamoDB
     from ScoutSuite.providers.aws.resources.config.service_private import Config
-    from ScoutSuite.providers.aws.services.dynamodb_private import DynamoDBConfig
     from ScoutSuite.providers.aws.services.kms_private import KMSConfig
 except ImportError:
-    ConfigConfig = None
-    DynamoDBConfig = None
+    Config = None
```
vifor2: Bye bye ConfigConfig :heart_eyes:

Issue #256, opened 2019-03-14
Managed policies count in IAM users report is wrong


Aboisier:
Managed policies count in IAM users report is wrong. I expected one, but it was zero.

![image](https://user-images.githubusercontent.com/6698529/54325024-92386c80-45d6-11e9-8375-c87ae2a45e73.png)

Aboisier:
Can we close this? Seems like it was imlemented in #277?

Issue #254, opened 2019-03-13
Inline policies in IAM groups are written weirdly big


Aboisier:
The image is pretty self-explanatory.

![image](https://user-images.githubusercontent.com/6698529/54314196-66ef5680-45b1-11e9-9736-54011897b59d.png)

Aboisier:
Can we close this? Seems like it was imlemented in #277?

Issue #253, opened 2019-03-13
Users do not appear in IAM groups


Aboisier:
When scanning IAM and viewing the groups in the report, the users' list is always empty, although the json is not.

![image](https://user-images.githubusercontent.com/6698529/54312830-483b9080-45ae-11e9-9c15-51f88e04267a.png)

Aboisier:
Can we close this, since it was imlemented in #277?

Issue #247, opened 2019-03-12
Fix broken collapsibles


vifor2:
There are quite a few collapsibles, at least on the AWS side that could be fixed or improved, here's a list of things to fix. Bare in mind that since I am working of a very fresh AWS account there aren't many ressources on it and therefore I am not testing all pages at the moment.

- [x] EC2 instances : the section *Network interfaces* is opened by default and pressing on the *Details* button causes it to close and then open almost instantly, afterwards it works as intended.
- [x] ELBv2 load balancers : the *Security groups* section has a button with the count of number of security groups in it, but clicking it launches an error in the console rather than making the section appear : `DOMException: Failed to execute 'querySelector' on 'Document': '#services.elbv2.regions.us-east-2.vpcs.vpc-d4f3****.lbs.5ca81197d3d5***********************************.security_groups' is not a valid selector.`
- [x] VPC vpcs : pressing on the number in the section *Network ACLs* does not make something appear, but does not throw an error either, is this intended? If so the button shouldn't suggest it is clickable (e.g. not change the cursor to a hand), else it should make the appropriate information appear.
- [x] S3 buckets : same issue as previous one for the sections *Groups with access via IAM policies* and *Roles with access via IAM policies*.

j4v:
While they appear to be buttons/collapsibles they aren't / shouldn't be, no? Those are just counters.

Also seeing that `Details` collapsible for AWS IAM Roles acts oddly in Chrome, sometimes you need to refresh for it to open/close.

Aboisier:
Can we close this? Seems like it was imlemented in #277?

PR #277, opened 2019-03-22
Enhancement/UI improvements/fixes

j4v:
Working on a few UI improvements. So far:

Modified https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/output/data/html/partials/accordion_policy.html so that it takes header size and `<samp></samp` as optional parameters, and adds `<code></code>` automatically.

Screenshot:
sc_2019-03-22_16h18m02s

Resolves https://github.com/nccgroup/ScoutSuite/issues/254

Correctly list IAM group user count and list members:

sc_2019-03-22_16h57m46s

Resolves https://github.com/nccgroup/ScoutSuite/issues/253

Correctly count Managed Policies:

sc_2019-03-22_17h08m12s

Resolves https://github.com/nccgroup/ScoutSuite/issues/256

Modified the appearance of the counter badges so that it's clear they aren't buttons (was quite confusing to people):

sc_2019-03-22_17h17m38s

Fix broken collapsible as described in https://github.com/nccgroup/ScoutSuite/issues/247:

sc_2019-03-22_19h03m58s

Latest commits should fix all occurrences of https://github.com/nccgroup/ScoutSuite/issues/247.

### DIFF
--- a/ScoutSuite/output/data/html/partials/accordion_policy.html
+++ b/ScoutSuite/output/data/html/partials/accordion_policy.html
@@ -1,11 +1,11 @@
     <!-- Policy partial -->
     <script id="accordion_policy.partial" type="text/x-handlebars-template">
       <div class="accordion">
-          <h4 class="list-group-item-heading accordion-heading"><samp>{{name}}</samp>
-          <a class="accordion-toggle collapsed" data-toggle="collapse" href="#{{escape_dots policy_path}}">
-            <span class="badge float-right btn-info">Details</span>
-          </a>
-        </h4>
+          <{{#if heading}}{{heading}}{{else}}h4{{/if}} class="list-group-item-heading accordion-heading">{{#samp}}<samp>{{/samp}}{{name}}{{#samp}}</samp>{{/samp}}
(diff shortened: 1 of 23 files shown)


j4v:
This should fix all the broken collapsibles as well as a number of UI bugs.

The only thing I wasn't able to solve was the S3 Bucket ACLs table which spills outside of the margin:

sc_2019-03-22_19h21m33s

@vifor2 could you have a look at that? I believe you've played with this code before.

j4v:
Also fixed the "Usage" output as raised in https://github.com/nccgroup/ScoutSuite/issues/12.

vifor2:
> The only thing I wasn't able to solve was the S3 Bucket ACLs table which spills outside of the margin:
> 
> sc_2019-03-22_19h21m33s
> 
> @vifor2 could you have a look at that? I believe you've played with this code before.

Yeah sure

vifor2:
Bucket name is now trimmed by default, user can click on the name to expand it and the overflow is viewable by using the horizontal scroll bar.
image

Issue #104, opened 2019-01-17
More user feedback when using the --service flag


Aboisier:
If the user writes an invalid service when using the `--service` flag (for example he/she writes `EC2` instead of `ec2`), all the services are scanned and there is no feedback.

It would be interesting to give the user some feedback. For example: 
> The `EC2` service does not exist. Here are the available services: ec2, cloudwatch, ...

j4v:
Agreed, and listing existing services would be great, it's not trivial to find the options at the moment.

PR #278, opened 2019-03-23
Fix for https://github.com/nccgroup/ScoutSuite/issues/104

j4v:
This is a fix for https://github.com/nccgroup/ScoutSuite/issues/104. 

I've put the path at the `ScoutSuite/providers/base/provider.py` level so it is dynamic/multi-provider. Ideally this should be done during argument parsing, but we don't know the list of available services at that time (without a hack).

Result:
sc_2019-03-23_19h10m21s


j4v:
Closing this one as tracked in https://github.com/nccgroup/ScoutSuite/pull/281 with a few fixes.

Issue #1, opened 2018-10-25
crash analytics github template


j4v:
When the application runs with errors, it should generate an anonymized GH-friendly template that can be pasted in a new issue and help remediation.

j4v:
Implemented in https://github.com/nccgroup/ScoutSuite/pull/281.

PR #281, opened 2019-03-24
Refactoring CLI Output

j4v:
This PR tracks a work in progress branch to refactor Scout's CLI output using coloredlogs.

Will look something like this:

sc_2019-03-24_19h56m25s

The implementation will be a bit less verbose then the current one as it will not output resource count (the "dynamic" count it fake anyway, and a list of resources can now be accessed through https://github.com/nccgroup/ScoutSuite/pull/279).

Instead of outputting stack traces during execution, these will be written to a file at the end of execution. This will also allow implementing https://github.com/nccgroup/ScoutSuite/issues/1.


j4v:
I think we can merge the current state of this PR even though it's still WIP @misg @Aboisier.

Normal run (for a refactored & a non-refactored service):

sc_2019-03-25_11h46m30s

Run with `--debug`, includes stacktrace:

sc_2019-03-25_11h47m24s

The exceptions (with additional context when provided) are written to disk:

sc_2019-03-25_11h46m41s

ScoutSuite/__init__.py
```diff
@@ -7,12 +7,14 @@
 EXCEPTIONS = 4242
 HTMLREPORT = 424242
 AWSRULESET = 42424242
+ERRORSLOG = 4242424242
```
Aboisier: I feel like the 42 thing won't scale very well. Maybe we should start using legit error codes hehe.

ScoutSuite/core/console.py
```diff
@@ -4,53 +4,68 @@
...
     global verbose_exceptions
     if verbose_exceptions:
-        print_error(str(traceback.format_exc()))
-    elif not debug_only:
-        print_error(str(e))
+        logger.exception(str)
+    else:
+        logger.error(str)
+
+    # logger.error('{} L{}: {}'.format(fname, exc_tb.tb_lineno, exception))
```
Aboisier: Why is this commented?

ScoutSuite/__init__.py
j4v: Well considering it's absurde origin (https://en.wikipedia.org/wiki/The_Hitchhiker%27s_Guide_to_the_Galaxy#The_Hitchhiker's_Guide_to_the_Galaxy)...

vifor2: :+1: for legit error codes pl0x

Aboisier: I am aware of the origin, great book, still think we should use legit error codes 😂

misg:
(The build was broken, I fixed the tests, there was a failing import because of this refactoring.)

ScoutSuite/__init__.py
j4v: Oh I completely agree, it's ridiculous. I'll do a bit of refactoring on this because it's not actually an error code / actually useful.

j4v:
@Aboisier got rid of the error codes as they were useless. Used it as an opportunity to make a few provider-agostic improvements.

@misg thanks for fixing the tests :fist_right:

j4v:
Merged https://github.com/nccgroup/ScoutSuite/pull/278:

sc_2019-03-26_10h10m16s

misg:
Would you need some help to resolve the conflicts with `refactoring/resource-configs`? Looks like there're a lot of them. :grimacing:

j4v:
> Would you need some help to resolve the conflicts with `refactoring/resource-configs`? Looks like there're a lot of them.

Should be OK now, there weren't any actual conflicts - I had branched out of an outdated local version of  `refactoring/resource-config`.

PR #286, opened 2019-03-27
Make the fetching of azure resources really concurrent

misg:
I just discovered that the azure python sdk that we use in our azure facades always return a custom iterator when the result of an API request is a collection of items. In fact, this custom iterator wraps a custom paginator that enables to fetch resources in a "lazy" way, making API calls only when necessary (e.g. only when truly iterating over that custom iterator).
Since our azure facades were always returning this kind of custom iterator, it means that all the API calls made for fetching the resources were actually done sequentially, outside the thread pool...

This PR fixes that, making the fetching of azure resources really concurrent (by simply enforcing iteration over those custom iterators, using python builtin `list()`).

Same PR applies to the [proprietary repository](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/119).

### DIFF
--- a/ScoutSuite/providers/azure/facade/keyvault.py
+++ b/ScoutSuite/providers/azure/facade/keyvault.py
@@ -7,4 +7,6 @@ class KeyVaultFacade:
         self._client = KeyVaultManagementClient(credentials, subscription_id)
 
     async def get_key_vaults(self):
-        return await run_concurrently(self._client.vaults.list_by_subscription)
+        return await run_concurrently(
+            lambda: list(self._client.vaults.list_by_subscription())
+        )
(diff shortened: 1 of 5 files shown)


j4v:
Nice find! What's the impact when the amount of resources is greater then what the iterator usually return? I imagine if there are thousands of resources, casting to a list might get out of hands?

misg:
> Nice find! What's the impact when the amount of resources is greater then what the paginator returns on each request? I imagine if there are thousands of resources, casting to a list won't return all the elements?

Azure SDK paginators are actually transparent to us, the SDK return custom iterators that enable us to iterate over all the results (from some elements to millions of elements). So in case of thousands/millions of resources, casting to a list will return all elements.
The main drawback I can see with this approach is that casting to a list when you got thousands (let's say more than `50 000`) of items will necessarily be slow (even though it will happened in a dedicated thread and not impact parallel fetching operations on other resources)... But It's definitely better than doing things sequentially and I think it's a good start to begin with and see how it will scale.

PR #288, opened 2019-03-27
Improve concurrent support for AWS services

Aboisier:
This PR aims to address the `TODO`s left during the refactoring (#183).


misg:
In the end, this PR mainly aims to improve concurrent support for all the fetching done for AWS services. This is done mainly by refactoring the code so to use the new function `get_and_set_concurrently` that easily enable to fetch and set concurrently all attributes related to a specific kind of resource.

Aboisier:
Approved,  @misg !

PR #289, opened 2019-03-27
Parallelize async loops and factorize async code

misg:
This PR aims to parallelize all the async loops that can be found in the current refactoring and factorize some async code, as part as fixing all TODOs of the current refactoring (#183).
To do so, a new method `_fetch_children_for_all_resources` has been introduced in AWSCompositeResources and AzureCompositeResources in order for a collection of composite resources to fetch all children of each resource in a concurrent way.


ScoutSuite/providers/aws/resources/resources.py
```diff
@@ -24,29 +23,49 @@ def __init__(self, facade, scope: dict):
...
     """
-        
-    async def _fetch_children(self, parent: object, scope: dict):
-        """This method calls fetch_all on each child defined in "_children" and stores the fetched resources \
-        in the parent under the key associated with the child. It also creates a "<child_name>_count" entry \
-        for each child.
 
-        :param parent: The object in which the children should be stored
-        :param scope: The scope passed to the children constructors
+    async def _fetch_children_of_all_resources(self, resources: dict, scopes: dict):
```
Aboisier: Should that be made _public_ (aka remove the `_`)? It seems like it's used outside of this class.

misg: Where? This method is always called from a child class of `AWSCompositeResources` (inside the `fetch_all` method), like `_fetch_children` and thus, never used outside of this class.

Aboisier: Oh you're right. I didn't realize they were all child classes. All good!

j4v:
gangster

PR #290, opened 2019-03-27
Refactoring/facade injection

Aboisier:
This PR aims to inject the facades in the different services instead of instantiating one in each service. This is related to #183 . Please also review the **[private pr](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/120)**.

SO sorry, it looks like a huge PR, but there are roughly 13 files you should really pay attention to, the rest is just redundant stuff:
* `__main__.py`
* `services.py` (all 4 files, for _aws_, _azure_, _gcp_ and the base one)
* `provider.py` (all 4 files)
* `authentication_strategy.py` (all 4 files)

## Here's what happens in this PR:
* The services won't need to know about the credentials since the facade will be injected already authenticated
* The providers' `authenticate` method was moved outside of the providers to `AuthenticationStrategy` classes.
* The authentication is now done in the `__main__` and the credentials are provided to the provider in the constructor
* This means that when the provider is instantiated, it is ready to be used (unlike the previous implementation where it had to be authenticated)

## Here's what didn't happen in this PR
* I did not inject the facade in the GCP service since they are not implemented yet
* Ideally, it should be possible to remove `credentials` from `Resource`'s `fetch_all` arguments. Currently, the credentials must be passed because of a few black sheep services. I didn't fix that because this PR was [big enough](https://www.youtube.com/watch?v=rvrZJ5C_Nwg).


ScoutSuite/providers/aws/resources/awslambda/service.py
```diff
@@ -17,8 +17,8 @@ def _parse_function(self, raw_function):
...
 class Lambdas(Regions):
     _children = [
-        (RegionalLambdas, 'functions')
+        (Functions, 'functions')
     ]
 
-    def __init__(self):
-        super(Lambdas, self).__init__('lambda')
+    def __init__(self, facade: AWSFacade):
+        super(Lambdas, self).__init__('lambda', facade)
```
Aboisier: Just thought I'd rename that while I was here, the name was bothering me.

j4v:
I really like the new authentication flow as it will allow failing gracefully on auth failure whereas previously we'd start trying to poll services even if we hadn't managed to authenticate.

PR #292, opened 2019-03-28
Refactoring/thread parameter

Aboisier:
This PR makes the number of threads in the thread pool parametrizable. I did not use the thread_config argument and decided to allow the user to specify the exact number of threads.

### DIFF
--- a/ScoutSuite/core/cli_parser.py
+++ b/ScoutSuite/core/cli_parser.py
@@ -203,11 +203,16 @@ class ScoutSuiteArgumentParser:
                             default=False,
                             action='store_true',
                             help='Do not automatically open the report in the browser.')
+        # TODO: This should be removed once the GCP refactoring is done
         parser.add_argument('--thread-config',
                             dest='thread_config',
                             type=int,
                             default=4,
                             help='Level of multi-threading wanted [1-5]; defaults to 4.')
+        parser.add_argument('--max-workers',
+                            dest='max_workers',
+                            type=int,
+                            help='Maximum number of threads used by Scout Suite')
         parser.add_argument('--report-dir',
                             dest='report_dir',
                             default=DEFAULT_REPORT_DIR,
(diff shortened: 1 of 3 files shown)


ScoutSuite/core/cli_parser.py
```diff
@@ -203,11 +203,16 @@ def _init_common_args_parser(self):
                             default=False,
                             action='store_true',
                             help='Do not automatically open the report in the browser.')
+        # TODO: This should be removed once the GCP refactoring is done
         parser.add_argument('--thread-config',
                             dest='thread_config',
                             type=int,
                             default=4,
                             help='Level of multi-threading wanted [1-5]; defaults to 4.')
+        parser.add_argument('--max-workers',
```
misg: Python documentation indicates:

> Changed in version 3.5: If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5, assuming that ThreadPoolExecutor is often used to overlap I/O instead of CPU work and the number of workers should be higher than the number of workers for ProcessPoolExecutor.

Reading your code, I guess default max_workers will be set to `0` and I'm not sure of the expected behavior... What do you think about adding a `default=None`?

Aboisier: If the argument is not provided by the user, the value is `None` by defaut.

image

misg: Perfect!

j4v:
> I did not use the thread_config argument and decided to allow the user to specify the exact number of threads.

I like the approach :medal_sports: , the old one was a bit obscure (how many threads does `5` run???). Also the number of threads you can run depends greatly on your platform.

my 2c:
- Do we want to set a default, e.g. 10? Current is 10 for listing, 20 for parsing (`4`).
- The arg uses "workers" but the description uses the term "threads", maybe that's a bit confusing without knowing the internals

PR #293, opened 2019-03-31
Make the resource fetching of all services concurrent

misg:
This PR brings the final enhancement about concurrency in our refactoring (#183) by fetching all the services of a cloud provider concurrently.
I tested it with an execution against AWS (all services, privates included) and it took 1min40 to build the report, instead of 2min30 for the version of Scout Suite on develop.
Definitely encouraging!

### DIFF
--- a/ScoutSuite/providers/base/configs/services.py
+++ b/ScoutSuite/providers/base/configs/services.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+import asyncio
 
 from ScoutSuite.utils import format_service_name
 from ScoutSuite.core.console import print_error, print_exception, print_debug, print_info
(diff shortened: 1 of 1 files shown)


ScoutSuite/providers/base/configs/services.py
```diff
@@ -16,28 +16,35 @@ def _is_provider(self, provider_name):
...
+        tasks = {
+            asyncio.ensure_future(
+                self._fetch(service, services, regions)
+            ) for service in vars(self)
+        }
+        await asyncio.wait(tasks)
+
+    async def _fetch(self, service, services, regions):
+        try:
+            # skip services
```
Aboisier: Est-ce qu'on pourrait skip les services dans `fetch` plutôt qu'ici? Ça éviterait d'avoir à passer `services` pour rien.

j4v:
> it took 1min40 to build the report, instead of 2min30 for the version of Scout Suite on develop

Is this using the same amount of threads/workers?

misg:
> > it took 1min40 to build the report, instead of 2min30 for the version of Scout Suite on develop
> 
> Is this using the same amount of threads/workers?

It was using 10 threads on our refactorized version and I thought it was the same for the version on `develop` but according to https://github.com/nccgroup/ScoutSuite/blob/develop/ScoutSuite/providers/aws/configs/regions.py#L134 it seems like it was actually using 20 threads (10 for fetching and 10 for parsing, by default (I didn't pass a `thread-config` argument))...

ScoutSuite/providers/base/configs/services.py
misg: Done!

Issue #240, opened 2019-03-11
Allow setting an arbitrary name for the report


j4v:
Currently the user can choose an arbitrary path, but not an arbitrary name.

Will help mitigate issues such as https://github.com/nccgroup/ScoutSuite/issues/233.

j4v:
This has been merged into `develop`.

Issue #154, opened 2019-02-07
Create report instance after authentication


j4v:
In `__main__.py`, the `report` object is created before authentication. The consequence is that we can't use authenticated data (e.g. GCP SA email, AWS account ID) to name the report (amongst other things we might want to do). 

Should move creation of `report` after authentication to leverage this info.

j4v:
This has been merged into `develop`.

PR #294, opened 2019-03-31
Make core provider-agnostic

j4v:
This PR ~~is a WIP which~~ aims to remove all mentions of "AWS" from the core, output and GCP/Azure code. 

The PR also updates a number of print/debug usages and renames occurrences of "scout2" to "scout".

The following are also implemented:
- https://github.com/nccgroup/ScoutSuite/issues/154
- https://github.com/nccgroup/ScoutSuite/issues/240


ScoutSuite/providers/aws/authentication_strategy.py
```diff
@@ -1,17 +1,20 @@
...
-    
+
     def authenticate(self, profile=None, **kwargs):
         session = boto3.Session(profile_name=profile)
         credentials = session.get_credentials().__dict__
 
         if credentials.get('access_key') is None:
-            raise AuthenticationException()
+            print_error('Failed to authenticate to AWS')
+            return False
```
misg: I think there's an inconsistency here with Azure and GCP `authenticate` methods. Both of them raise an `AuthenticationException` in case of authentication failure and don't return a `False` value...  But I don't really know if we want to return a `False` value or raise an exception, to be honest. Any opinion on the matter @Aboisier ?

j4v: Yeah I wasn't sure either. I think returning False and printing an exception would be the best option, as it would allow `__main__.py` to exit graciously. No real reason to throw a custom exception, no?

ScoutSuite/core/conditions.py
```diff
@@ -33,7 +30,7 @@ def pass_conditions(all_info, current_path, conditions, unknown_as_pass_conditio
         return True
     condition_operator = conditions.pop(0)
     for condition in conditions:
-        if condition[0] in condition_operators:
+        if condition[0] in ['and', 'or']:
```
Aboisier: I'm curious, why this change?

ScoutSuite/providers/aws/authentication_strategy.py
Aboisier: I'm not a fan of using a boolean to indicate success or failure when failure is not a valid use case. In this context, failure is truly exceptional, and it's important to let the caller know that it failed. The caller should handle the exception and log if it wants. The boolean solution is less verbose and could lead to problems if this method is used elsewhere.

ScoutSuite/providers/gcp/utils.py
```diff
@@ -39,7 +38,7 @@ def gcp_connect_service(service, credentials=None, region_name=None):
         if service == 'stackdriverlogging':
             # return stackdriver_logging.LoggingServiceV2Client()
             return stackdriver_logging.Client(project='placeholder')  # need a project value to instantiate the client
-                                                                      # event though it won't be used afterwards
+            # event though it won't be used afterwards
```
Aboisier:
```suggestion
            # even though it won't be used afterwards
```

ScoutSuite/providers/gcp/utils.py
```diff
@@ -39,7 +38,7 @@ def gcp_connect_service(service, credentials=None, region_name=None):
         if service == 'stackdriverlogging':
             # return stackdriver_logging.LoggingServiceV2Client()
```
Aboisier:
```suggestion
```

ScoutSuite/providers/aws/authentication_strategy.py
misg: I think I agree with @Aboisier on this one.

ScoutSuite/core/conditions.py
j4v: `condition_operators` was the only variable referenced in https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/core/__init__.py (the rest was dead code), and it was only referenced at that line in `conditions.py`, figured I'd remove the file as part of the clean up (less is more)

j4v:
@misg & @Aboisier I've normalized the authentication error handling in https://github.com/nccgroup/ScoutSuite/pull/294/commits/f032971c1dfd41a77efa093ed635873bc30a7c00, doesn't rely on a boolean anymore.

ScoutSuite/core/conditions.py
Aboisier: Ok, nice! 💯

PR #302, opened 2019-04-03
Refactoring/resource configs cleanup

Aboisier:
This PR is part of #183 . It cleans up the refactoring a bit. For example, there was a lot of redundant file names. For example, `facade/facade.py` was renamed to `facade/__init__.py`. This simplifies the import. Some dead code was removed too.


j4v:
:clap: :clap: :clap: 

As part of this could you remove the `providers/<provider>/services/` directories (in public and private repos)?

ScoutSuite/providers/base/configs/base.py
```diff
@@ -83,40 +81,23 @@ def get_non_provider_id(self, name):
...
-            else:
-                api_client = connect_service(self.service, credentials, silent=True)
+            print_error('You should not use BaseConfig for AWS services. Please refer to the wiki: https://bit.ly/2IbiWtA.')
 
         elif self._is_provider('gcp'):
             api_client = gcp_connect_service(service=self.service, credentials=credentials)
 
         elif self._is_provider('azure'):
-            api_client = azure_connect_service(service=self.service, credentials=credentials)
+            print_error('You should not use BaseConfig for Azure services. Please refer to the wiki: https://bit.ly/2IbiWtA.')
```
vifor2: Ohhhh even minified the link, neato

Aboisier: Of course, I ain't a barbarian man.

Remi05:
I'm not a fan of putting classes in `__init__.py`. I'm pretty sure that by convention, only package-related initialisations and imports should be done in it, but no more. You can keep the cleaner imports by importing the classes you want to expose in your `__init__.py`.

ie.
In `ScoutSuite/providers/aws/resources/ec2/__init__.py`, you would have :
```
from .service import EC2
```
This would allow to import `EC2` with `from ScoutSuite.providers.aws.resources.ec2 import EC2` instead of `from ScoutSuite.providers.aws.resources.ec2.service import EC2` while also keeping `__init__.py` clean. 

You can find more info on packages here :
https://docs.python.org/3/tutorial/modules.html#packages

Let me know what you all think!

Aboisier:
> I'm not a fan of putting classes in `__init__.py`. I'm pretty sure that by convention, only package-related initialisations and imports should be done in it, but no more. You can keep the cleaner imports by importing the classes you want to expose in your `__init__.py`.
> 
> ie.
> In `ScoutSuite/providers/aws/resources/ec2/__init__.py`, you would have :
> 
> ```
> from .service import EC2
> ```
> This would allow to import `EC2` with `from ScoutSuite.providers.aws.resources.ec2 import EC2` instead of `from ScoutSuite.providers.aws.resources.ec2.service import EC2` while also keeping `__init__.py` clean.
> 
> You can find more info on packages here :
> https://docs.python.org/3/tutorial/modules.html#packages
> 
> Let me know what you all think!

I don't see any reason why we can't put classes in `__init__.py`, except for personal preferences... I skimmed through the convention, it doesn't seem like it is prohibited. Some official packages define methods in `__init__.py` (see [this package](https://github.com/python/cpython/blob/3.7/Lib/json/__init__.py), for example). I understand it is not a common practice, but I feel like as long as we're consistent, it's fine. It makes it easier to navigate the resources folders, simplifies the imports, and doesn't require adding another to the process of implementing a service.

Remi05:
> > I'm not a fan of putting classes in `__init__.py`. I'm pretty sure that by convention, only package-related initialisations and imports should be done in it, but no more. You can keep the cleaner imports by importing the classes you want to expose in your `__init__.py`.
> > ie.
> > In `ScoutSuite/providers/aws/resources/ec2/__init__.py`, you would have :
> > ```
> > from .service import EC2
> > ```
> > This would allow to import `EC2` with `from ScoutSuite.providers.aws.resources.ec2 import EC2` instead of `from ScoutSuite.providers.aws.resources.ec2.service import EC2` while also keeping `__init__.py` clean.
> > You can find more info on packages here :
> > https://docs.python.org/3/tutorial/modules.html#packages
> > Let me know what you all think!
> 
> I don't see any reason why we can't put classes in `__init__.py`, except for personal preferences... I skimmed through the convention, it doesn't seem like it is prohibited. Some official packages define methods in `__init__.py` (see [this package](https://github.com/python/cpython/blob/3.7/Lib/json/__init__.py), for example). I understand it is not a common practice, but I feel like as long as we're consistent, it's fine. It makes it easier to navigate the resources folders, simplifies the imports, and doesn't require adding another to the process of implementing a service.

I know it's not prohibited but the logic I have usually encountered is for it to be used for things you want to be available package-wide (ie. functions, variables, imports, etc.). I fear that other contributors might find it odd for the classes we would be putting there to be in this file. Though, I guess it mainly comes down to @j4v 's and the rest of the team's preference!

j4v:
> I know it's not prohibited but the logic I have usually encountered is for it to be used for things you want to be available package-wide (ie. functions, variables, imports, etc.). I fear that other contributors might find it odd for the classes we would be putting there to be in this file. Though, I guess it mainly comes down to @j4v 's and the rest of the team's preference!

After reading up on it and looking at the changes I'm not too keen on having the base class definitions in `__init__.py`. It's true that we *can* do it, but convention is really to put class initializations rather than definitions in there. As most of our external contributions are one-offs, I'd rather if we stick to general python conventions.

There are also some changes since 3.3 to how python uses the init files https://stackoverflow.com/questions/37139786/is-init-py-not-required-for-packages-in-python-3, we could get rid of a number of them.

What I propose are the following changes:
- Move the `<provier>Facade` class definitions to `/facade/base.py`
  -  For AWS couldn't we get rid of `AWSBaseFacade` (and thus `facade/basefacade.py`)? Doesn't seem to implement much
- Rename `/resources/resources.py` to `/resources/base.py`
  - Some of that code will be moved up to the `base` provider once GCP is done I'd hope
- For each resource folder, put the "main" class in `/resources/<service>/base.py`
- Move all utils to `/providers/<provider>/utils.py`, because e.g. for Azure there are 2 `utils.py` files which pretty much serve the same purpose

Your thoughts? I think that would keep the hierarchy consistent between providers and make code navigation/discovery easier.
We're getting real close to a :100: architecture for the providers, this is really "avant-garde" work :1st_place_medal:.

j4v:
It also looks like the file hierarchy of the `base` provider doesn't match that of the refactored providers? Most of the files are in the `configs` folder, which no longer exists.

I think there might also be some dead/legacy code in there, e.g. the `ScoutSuite/providers/base/configs/threads.py` file and some code in `ScoutSuite/providers/base/configs/base.py` (not sure about that one).

Aboisier:
Democracy wins!


> It also looks like the file hierarchy of the `base` provider doesn't match that of the refactored providers? Most of the files are in the `configs` folder, which no longer exists.
> 
> I think there might also be some dead/legacy code in there, e.g. the `ScoutSuite/providers/base/configs/threads.py` file and some code in `ScoutSuite/providers/base/configs/base.py` (not sure about that one).

I'm waiting for the GCP refactoring to be done to address those. 👍

PR #311, opened 2019-04-07
Azure cleanup

misg:
This PR aims to do some cleanup in Azure support, as part of the refactoring (#183).
In particular, the `scope` concept used in AWS support has been introduced (that makes things clearer), a class `AzureResources` has been defined to make `facade` and `scope` root attributes (instead of passing them as constructor parameters at each level), resource count computations have been moved to the `_fetch_children` method instead of being explicitly done in every resource classes and finally, the storage account facade has been factorized using `get_and_set_concurrently` helper.
Same improvements should be coming for Azure private services asap.


ScoutSuite/providers/azure/resources/base.py
```diff
@@ -3,11 +3,26 @@
...
-class AzureCompositeResources(CompositeResources, metaclass=abc.ABCMeta):
+class AzureResources(Resources, metaclass=abc.ABCMeta):
+
+    """This is the base class for Azure resources."""
+
+    def __init__(self, facade: AzureFacade, scope: dict = None):
+        """
+        :param scope: The scope holds the scope in which the resource is located. This usually means at least a region,
+                      but can also contain a VPC id, an owner id, etc. It should be used when fetching the data through
+                      the facade.
```
Aboisier: Does this comment apply to Azure? I thought there weren't regions with Azure. I might very well be wrong.

zer0x64: Pretty sure that's a copy paste error, I never saw Owner and VPCs in azure!

misg: Indeed, I updated the docstring. Nice catch. :)

PR #312, opened 2019-04-08
Enhancement/Additional AWS exception handling

j4v:
After running Scout against a production account for which I didn't have sufficient IAM privileges, I realized that **none** of the exceptions were being handled (which resulted in an empty errors file).

I've added exception handling to all the API calls made in the facades and a few more files.

@misg & @Remi05 can you please ensure something similar is implemented for Azure / GCP? It's pretty easy to test, just run scout with a user that doesn't have any privileges and see where it fails!
I just looked for all the instances of an API client and added try/catch clauses.

### DIFF
--- a/ScoutSuite/providers/aws/resources/directconnect/base.py
+++ b/ScoutSuite/providers/aws/resources/directconnect/base.py
@@ -6,9 +6,10 @@ from ScoutSuite.providers.aws.resources.base import AWSResources
 class Connections(AWSResources):
     async def fetch_all(self, **kwargs):
         raw_connections = await self.facade.directconnect.get_connections(self.scope['region'])
-        for raw_connection in raw_connections:
-            name, resource = self._parse_function(raw_connection)
-            self[name] = resource
+        if raw_connections:
+            for raw_connection in raw_connections:
+                name, resource = self._parse_function(raw_connection)
+                self[name] = resource
 
     def _parse_function(self, raw_connection):
         raw_connection['id'] = raw_connection.pop('connectionId')
--- a/ScoutSuite/providers/aws/resources/ec2/ami.py
+++ b/ScoutSuite/providers/aws/resources/ec2/ami.py
@@ -4,9 +4,10 @@ from ScoutSuite.providers.aws.resources.base import AWSResources
 class AmazonMachineImages(AWSResources):
     async def fetch_all(self, **kwargs):
         raw_images = await self.facade.ec2.get_images(self.scope['region'], self.scope['owner_id'])
-        for raw_image in raw_images:
-            name, resource = self._parse_image(raw_image)
-            self[name] = resource
+        if raw_images:
+            for raw_image in raw_images:
+                name, resource = self._parse_image(raw_image)
+                self[name] = resource
 
     def _parse_image(self, raw_image):
         raw_image['id'] = raw_image['ImageId']
--- a/ScoutSuite/providers/aws/resources/vpcs.py
+++ b/ScoutSuite/providers/aws/resources/vpcs.py
@@ -1,3 +1,4 @@
+from ScoutSuite.core.console import print_exception
 from ScoutSuite.providers.aws.resources.base import AWSCompositeResources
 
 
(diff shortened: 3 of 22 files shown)


ScoutSuite/providers/base/provider.py
```diff
@@ -365,19 +363,29 @@ def _new_go_to_and_do(self, current_config, path, current_path, callbacks):
...
-                            if type(current_config[key] == dict) and type(value) != dict and type(value) != list:
-                                callback(current_config[key][value],
-                                         path,
-                                         current_path,
-                                         value,
-                                         callback_args)
-                            else:
-                                callback(current_config, path, current_path, value, callback_args)
+                            try:
+                                # callback = globals()[callback_name]
```
Aboisier:
```suggestion
```

j4v:
> Can't believe none of us thought of that before

It didn't take *too* long to implement. It's really the facades' job to ensure exceptions are handled and something is returned in every case (e.g. an empty list if the call failed, with an error logged with `print_exception`).

ScoutSuite/providers/utils.py
```diff
@@ -45,12 +46,15 @@ def run_concurrently(func):
...
-    }
-    await asyncio.wait(tasks)
+    try:
+        tasks = {
+            asyncio.ensure_future(
+                get_and_set_func(entity, **kwargs)
+            ) for entity in entities for get_and_set_func in get_and_set_funcs
+        }
+        await asyncio.wait(tasks)
+    except Exception as e:
```
misg: I'm afraid that with this `try... except` clause, any exception that could occured during the execution of `func` will be catched and thus, won't be propagated any further.
It would mean that all of the exception handling done above for AWS facades will never be used and that in case of failures during data fetching, the only kind of logs you would get would be `Failed to run function concurrently: ...`. 

Or did you test it and I'm completely wrong?

ScoutSuite/providers/aws/resources/directconnect/base.py
```diff
@@ -6,9 +6,10 @@
 class Connections(AWSResources):
     async def fetch_all(self, **kwargs):
         raw_connections = await self.facade.directconnect.get_connections(self.scope['region'])
-        for raw_connection in raw_connections:
-            name, resource = self._parse_function(raw_connection)
-            self[name] = resource
+        if raw_connections:
```
misg: I think this is useless since `get_connections` never returns `None`. If `get_connections` fails, it returns an empty list and thus, iterating over it works as intented. What's the point of adding a condition like that, that adds a new level of indentation and introduce a git diff of 7 lines?

ScoutSuite/providers/utils.py
```diff
@@ -68,13 +72,16 @@ def run_concurrently(func):
...
+    try:
+        tasks = {
+            asyncio.ensure_future(
+                coro(entity, **kwargs)
+            ) for entity in entities
+        }
+        for task in asyncio.as_completed(tasks):
+            result = await task
+            results.append(result)
+    except Exception as e:
```
misg: Same thing as above for `run_concurrently`.

ScoutSuite/providers/aws/resources/ec2/ami.py
```diff
@@ -4,9 +4,10 @@
 class AmazonMachineImages(AWSResources):
     async def fetch_all(self, **kwargs):
         raw_images = await self.facade.ec2.get_images(self.scope['region'], self.scope['owner_id'])
-        for raw_image in raw_images:
-            name, resource = self._parse_image(raw_image)
-            self[name] = resource
+        if raw_images:
```
misg: Same as above.

ScoutSuite/providers/aws/resources/vpcs.py
```diff
@@ -12,16 +13,25 @@ def __init__(self, facade, scope: dict, add_ec2_classic=False):
...
-        for raw_vpc in raw_vpcs:
-            vpc_id, vpc = self._parse_vpc(raw_vpc)
-            self[vpc_id] = vpc
 
-        await self._fetch_children_of_all_resources(
-            resources=self,
-            scopes={vpc_id: {'region': self.scope['region'], 'vpc': vpc_id}
-                    for vpc_id in self}
-        )
+        try:
```
misg: Exception handling is already done in EC2 facade (thanks to your modifications above), what's the point of doing it again here?

misg:
I'm a bit late for my review, unfortunately. I agree with all the modifications you did in AWS facades, they are really a nice enhancement! (I'll do the same for Azure.) :100: Although, I think I disagree with the few modifications you did in the resource files, as well with those done in `utils.py`. Please take a look at my comments.

PR #316, opened 2019-04-09
Manual merge/develop 226

vifor2:
Feature functional although it could use some improvements (hence the EXPERIMENTAL disclaimer when using `--help`), but since there is less than 1 week before the end of the project @zer0x64 and me would like to get this merged ASAP! Also to cease the merge conflicts (**pl0x**).

[This PR](https://github.com/nccgroup/ScoutSuite/pull/260) but less chaotic.
Implements [this issue](https://github.com/nccgroup/ScoutSuite/issues/226).

## Scroll down for TL;DR

Changes done within this PR :
- User can input `--result-format` with either `json` or `sqlite` as args.
- Imports of `scoutsuite-results/scoutsuite_results.js` and `scoutsuite-results/scoutsuite_exceptions.js` are no longer in `report.html` by default, they are instead in a partial that is added if the user chose `json` in the previous point.
- `scoutsuite.js` now checks whether an element with the id `json_format` or `sqlite_format` exists to know which functions to call.
- When choosing SQLite all of the user's data is saved as a `.db` file instead of a `.js` file, these files being more compact as well as readable by a server.
- User can input arguments related to the server (reminder: we have decided that when the user selects SQLite he must then enter a second command to run the server, it will not be launched automatically).
- Exposed all of the data saved in the `.db` file to our client through **CherryPy**.
- Modified all `.js` files to follow [standardJS](https://standardjs.com/), there are still some changes to do.
- All data except resource pages other than the first one are fetched at the start and properly viewable within the report.
- Added **Back** and **Next** buttons to change page, pressing one of those buttons disables it if you're at the page limit afterwards ([0, totalResources / pageSize - 1]), it removes the data of the previous resource from memory (run_results) and add the new page to it, it then overwrites the current template with the same one but with the new data within it.
- Page size is currently hard-coded it would be nice to give the option to the user of choosing the size but Id rather get this issue done and merged beforehand. Same with more buttons to go to first page, last page, page at a specific index, modal that tells the user to make sure his port 8000 is freed and the server is running, etc.
- I know that when we talked I said Id keep the paging buttons on both result formats but for now the buttons are disabled when viewing a report in JSON format (it should look as if nothing had happened) while the SQLite one have 2 buttons.
- Merged develop and it's new architecture into this branch.

Known issues : 
- Works with **Firefox** but not with **Chrome**, will discuss about this issue tomorrow with @j4v : `jquery-3.3.1.min.js:2 Access to XMLHttpRequest at 'file:///home/vifor2/Documents/ScoutSuite/scoutsuite-report/inc-scoutsuite/sqlite.js' from origin 'null' has been blocked by CORS policy: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.`
- Items in the navbar are not in alphabetical order although they are in the JavaScript object (I'm bamboozled).
- Constant error when used with **Firefox** although it doesn't seem to actually cause any problem : `XML Parsing Error: syntax error
Location: file:///home/vifor2/Documents/ScoutSuite/scoutsuite-report/inc-scoutsuite/sqlite.js
Line Number 1, Column 1:`
- Initial data loading is a bit long compared to other providers because of the 16 regions per resource.

### Example with Azure
While in /ScoutSuite open two terminals, in the first one type in `azure --cli --serve` and in the second one your usual command followed by `--result-format sqlite`, you can then view the report has usually excepted for the fact that you'll have buttons to load in/out new pages.
### Example with GCP
Same as concept as Azure but replace the first command with `gcp --user-account --serve`
### Example with AWS
Same concept as previous ones but replace the first command with `aws --serve`

Or you can only use one terminal, `--no-browser` and the 2 commands in the right order...


ScoutSuite/core/cli_parser.py
```diff
@@ -239,9 +235,32 @@ def _init_common_args_parser(self):
...
+                            nargs='+',
                             help='Exception file to use during analysis.')
+        parser.add_argument('--result-format',
+                            dest='result_format',
+                            default='json',
+                            type=str,
+                            choices=['json', 'sqlite'],
+                            help="[EXPERIMENTAL FEATURE] The database file format to use. JSON doesn't require a server to view the report, "
+                                 "but cannot be viewed if the result file is over 400mb.")
+        parser.add_argument('--serve',
```
Aboisier: Can we put the server related stuff in a sub-parser or something like that?

Copied from #260, @zer0x64 :
> This is kinda of a touchy subject, because
>
> 1. Theorically, the subparsers are meant to be providers
I understand it's currently the case, but I don't think we ever said subparsers were meant to be only for providers :o 

> 2. The way it is done right now, because of the execution order, you need to log in before being able to serve the report. This would require some heavier changes for now than what it's actually worth, because it touches how to provider object is created

I guess we can create an issue if it implies too many changes, but the `--serve` argument doesn't really makes sense in my opinion, it's very confusing.

ScoutSuite/core/server.py
```diff
@@ -0,0 +1,107 @@
+from sqlitedict import SqliteDict
```
Aboisier: Copied from #260 , @Aboisier :
> This class lacks docstrings, and the method names are also very generic. It's a bit hard to understand what is going on. I think unit tests would be very appropriate and easy to implement.

ScoutSuite/core/server.py
```diff
@@ -0,0 +1,107 @@
...
+        if not isinstance(data, dict):
+            return data
+
+        result = {}
+        for k, v in data.items():
+            if isinstance(v, dict):
+                result[k] = {'type': 'dict', 'keys': list(v.keys())}
+            elif isinstance(v, list):
+                result[k] = {'type': 'list', 'length': len(v)}
+        return result
```
Aboisier: Copied from #260, @Aboisier :
> Are we actually using that or are we still looking for a solution haha? One would be to escape the keys and then use dots.

ScoutSuite/output/data/html/partials/aws/services.ec2.regions.vpcs.security_groups.resource_list.html
```diff
@@ -14,7 +14,7 @@ <h5 class="list-group-item-heading accordion-heading">{{ make_title service }} {
                   <li class="list-group-item-text">
                     <a href="javascript:showObject('services.{{../service}}.regions.{{../region}}.vpcs.{{../vpc}}.{{../resource_type}}.{{this}}')">
                         {{this}}
-                        <!--{{get_value_at 'services' ../service 'regions' ../region 'vpcs' ../vpc ../resource_type this 'name'}}-->
+                        <!--{{getValueAt 'services' ../service 'regions' ../region 'vpcs' ../vpc ../resource_type this 'name'}}-->
```
Aboisier:
```suggestion
```

zer0x64:
@Aboisier Yeah, CORS crashing on chrome is a known issue(we talked about it with @j4v last week). Nice job finding the flag, we'll add it to the doc!

ScoutSuite/core/cli_parser.py
zer0x64: Will open an issue.

ScoutSuite/core/server.py
zer0x64: Not really sure unit testing would be easy as it requires a prior run for data, implies an SQLite database and an HTTP server which can be a real pain to test because of latency and would require a new dependency for testing.

I will get on the docstrings right away!

ScoutSuite/core/server.py
zer0x64: I don't see a problem using that :)

ScoutSuite/core/server.py
zer0x64: Done!

ScoutSuite/providers/base/provider.py
```diff
@@ -25,7 +25,7 @@ class BaseProvider(object):
     """
 
     def __init__(self, report_dir=None, timestamp=None, services=None, skipped_services=None, thread_config=4,
-                 **kwargs):
+                 result_format='json', **kwargs):
```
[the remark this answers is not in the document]
zer0x64: That's @vifor2 works.
That might have to do with the report generation, so it ends up written somewhere in the javascript or HTML that the report is supposed to be hooked to a server. I see it stored in the provider object at least in Azure, though I don't know if it's used anywhere, and in the base provider it is not used anywhere. Might just be leftovers

ScoutSuite/output/result_encoder.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    JSON encoder class
+    """
+
+    def default(self, o):
+        try:
+            if type(o) == datetime.datetime:
+                return str(o)
+            else:
+                # remove unwanted attributes from the provider object during conversion to json
```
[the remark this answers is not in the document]
zer0x64: That will be the job of the last PR to be merged, so we don't break everything yet :)

ScoutSuite/providers/base/provider.py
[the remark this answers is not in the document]
vifor2: Yeah depeding on what type of report I attach or not certain elements to the report, e.g. the script sqlite.js is not attached if creating a json report. This also helps me determine which functions to run in scoutsuite.js

ScoutSuite/output/data/inc-scoutsuite/pagination.js
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+      for (let resource in runResults['services'][service]) {
+        if (resource.match(reCount)) {
+          let pathArray = ['services', service, resource.replace(reCount, '')]
+          loadPage(pathArray, 0)
+        }
+      }
+    }
+  }  
+}
```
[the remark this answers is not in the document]
vifor2: Sure I'll fragment it, it'll end up going to the same depth though

zer0x64:
> I'm getting this error log when scanning.
> 
> ```
> 2019-04-09 10:47:24 SURFACE-ANTOINE scout[6792] INFO Applying exceptions
> 2019-04-09 10:47:24 SURFACE-ANTOINE scout[6792] ERROR scout.py L186: Failed to load exceptions: 'NoneType' object has no attribute 'replace'
> ```

This is on develop too, and happens when there are exceptions(to enter the if execptions:) and the profile is None, so on both GCP and Azure.

ScoutSuite/output/data/inc-scoutsuite/pagination.js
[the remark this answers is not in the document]
vifor2: Will fragment!

ScoutSuite/output/data/inc-scoutsuite/sqlite.js
```diff
@@ -0,0 +1,109 @@
...
+  // Save the current page index to remember which page we have saved
+  // Originally wanted to save that info under the precise resource, but the handlebar templates create slots for
+  // each entry under resource, therefore there were 2 empty slots always added
+  runResults['services'][service][resource + '_page_index'] = pageIndex
+  runResults['services'][service][resource + '_page_size'] = pageSize
+}
+
+/**
+ * Acts like getResourcePageSqlite but when we're using regions, made a separate function since the order of
+ * the variables are different and it was getting confusing
```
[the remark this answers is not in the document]
vifor2: Made separate functions to limit the amount of if/elsr clauses. But yeah, more needs to be done for sure.

ScoutSuite/providers/base/provider.py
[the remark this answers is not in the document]
Aboisier: What does the provider have to do with this though? :/

ScoutSuite/output/data/inc-scoutsuite/sqlite.js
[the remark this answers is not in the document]
Aboisier: I meant why do we need specific pieces of code for both cases. The comment is a bit unclear, what variables are different? And how are the _projects_ handled?

ScoutSuite/core/server.py
Aboisier: If the code is not testable, then there is a problem :|

j4v:
> This is on develop too, and happens when there are exceptions(to enter the if execptions:) and the profile is None, so on both GCP and Azure.

Should be fixed in https://github.com/nccgroup/ScoutSuite/pull/327.

PR #319, opened 2019-04-09
Refactoring/gcp/gce

Remi05:
Migrated Google Compute Engine to the new architecture. Also adds support for filtering GCE Instances by Zone and GCE Subnetworks by Region.


ScoutSuite/providers/gcp/resources/gce/firewalls.py
```diff
@@ -0,0 +1,59 @@
...
+        firewall_dict['network'] = raw_firewall['network'].split('/')[-1]
+        firewall_dict['network_url'] = raw_firewall['network']
+        firewall_dict['priority'] = raw_firewall['priority']
+        firewall_dict['source_ranges'] = raw_firewall.get('sourceRanges', [])
+        firewall_dict['source_tags'] = raw_firewall.get('sourceTags', [])
+        firewall_dict['target_tags'] = raw_firewall.get('targetTags', [])
+        firewall_dict['direction'] = raw_firewall['direction']
+        firewall_dict['disabled'] = raw_firewall['disabled']
+
+        # Parse FW rules
```
Remi05: Clean this up / move it to a separate method.

ScoutSuite/providers/gcp/resources/gce/instances.py
```diff
@@ -0,0 +1,73 @@
...
+        return description if description else 'N/A'
+
+    def _is_block_project_ssh_keys_enabled(self, raw_instance):
+        metadata = self._metadata_to_dict(raw_instance['metadata'])
+        return metadata.get('block-project-ssh-keys') == 'true'
+
+    def _metadata_to_dict(self, metadata):
+        return dict((item['key'], item['value']) for item in metadata['items']) if 'items' in metadata else {}
+
+    async def _get_common_instance_metadata_dict(self):
```
misg: This method should be moved to GCE facade and those metadata should get attached to their instance objects when calling `get_instances` in order to maintain the separation of concerns principle (between the fetching and parsing parts) and also avoid performance issues (in this case, since the parsing is `async` because it does some fetching, the `for loop` in `fetch_all` is not really efficient and could lead to poor performances).
Although, regarding our schedule, I don't mind (do you @Aboisier ?) if you simply add a `TODO` to move forward and merge those PRs so we can unify a last time our refactoring.

ScoutSuite/providers/gcp/resources/gce/service.py
```diff
@@ -0,0 +1,32 @@
...
+        # and Zones classes, but having a "resource_name_count" field in their 
+        # dictionary causes errors in the rule engine since they are accessed
+        instances_count = 0
+        subnetworks_count = 0
+        for project in self['projects'].values():
+            for zone in project['zones'].values():
+                instances_count += zone['instances_count']
+            for region in project['regions'].values():
+                subnetworks_count += region['subnetworks_count']
+        self['instances_count'] = instances_count
```
misg: I think a one liner would be more pythonic (don't forget to remove lines 24 to 30 if you agree with that):
```suggestion
        self['instances_count'] = sum([zone['instances_count'] for zone in project['zones'].values() for project in self['projects'].values()])
```

ScoutSuite/providers/gcp/resources/gce/service.py
```diff
@@ -0,0 +1,32 @@
...
+        self['subnetworks_count'] = subnetworks_count
```
misg: Same here:
```suggestion
        self['subnetworks_count'] = sum([region['subnetworks_count'] for region in project['regions'].values() for project in self['projects'].values()])
```

Aboisier:
@remi05 is this review-ready?

ScoutSuite/providers/gcp/provider.py
```diff
@@ -87,7 +87,7 @@ def preprocessing(self, ip_ranges=None, ip_ranges_name_key=None):
         """
         ip_ranges = [] if ip_ranges is None else ip_ranges
 
-        self._match_instances_and_snapshots()
+        # self._match_instances_and_snapshots()
```
Aboisier: What's up with that?

ScoutSuite/providers/gcp/resources/gce/instances.py
Aboisier: It would be ideal is if was fixed before merging, but I understand we're running out of time now. :/

ScoutSuite/output/data/html/partials/gcp/left_menu_for_gcp_region.html
```diff
@@ -0,0 +1,38 @@
+<!-- Generic left-menu link partial for region-then-resource -->
+<script id="left_menu_for_gcp_region.partial" type="text/x-handlebars-template">
+  <div id="services.{{service_name}}.projects.id.regions.id.{{resource_type}}.list">
+  <div class="list-group" >
+    <div class="list-group-item">
+      <a href="javascript:showRowWithItems('services.{{service_name}}.projects.id.regions.id.{{resource_type}}')">Show all <span class="badge float-right btn-info">{{get_value_at 'metadata' service_group service_name 'resources' resource_type 'count'}}</span></a>
```
Aboisier: `get_value_at` was renamed, it's now `getValueAt` I believe.

Aboisier:
> I'm also wondering if the `Regions` and `Zones` classes you defined couldn't be moved "up" (to `providers/gcp/resources/`, at the same level than `GCPCompositeResources`) in order to be "abstract" (like AWS `Regions` and `VPCs`) and be used for potential other GCP services. What do you think @Aboisier ?

Yup, looks like that would make sense! Ì haven't worked with GCP enough to know if there are other services which use regions. I'm guessing there are.

ScoutSuite/providers/gcp/resources/gce/service.py
Remi05: Sure, I was mainly looking to avoid super long lines tbh

Remi05:
> > I'm also wondering if the `Regions` and `Zones` classes you defined couldn't be moved "up" (to `providers/gcp/resources/`, at the same level than `GCPCompositeResources`) in order to be "abstract" (like AWS `Regions` and `VPCs`) and be used for potential other GCP services. What do you think @Aboisier ?
> 
> Yup, looks like that would make sense! Ì haven't worked with GCP enough to know if there are other services which use regions. I'm guessing there are.

I tried to do that at first but it caused a lot of issues, it's a bit different than AWS' VPC and Region. Anyway, the Regions and Zones are currently only used in GCE's API (to fetch regions and zones, we use the Compute Engine client).

ScoutSuite/output/data/inc-scoutsuite/scoutsuite.js
```diff
@@ -97,6 +98,12 @@ function onPageLoad() {
 
 };
 
+function registerHandlebarHelpers() {
+    Handlebars.registerHelper('concat', function(str1, str2) {
+        return [str1, str2].join('');
+    });
+}
+
```
Aboisier: There is already a bunch of helpers registered, and I'm pretty sure there is one for concatenation... why did you add `registerHandlebarHelpers`? 

Lemme find that helper

Remi05: I didn't notice the existing ones but the current `concat` joins the string with a dot (ie. `concat str1 str2` returns `str1.str2` instead of `str1str2`). I'll move mine to the `helpers.js` file and rename it.

ScoutSuite/providers/gcp/resources/gce/instances.py
Remi05: Good idead! I wasn't a fan of having an `async` parsing method.

ScoutSuite/providers/gcp/provider.py
Remi05: Forgot to uncomment it, good catch!

ScoutSuite/providers/gcp/resources/gce/instances.py
```diff
@@ -0,0 +1,73 @@
...
+    _children = [ 
+        (InstanceDisks, 'disks')
+    ]
+
+    def __init__(self, gcp_facade, project_id, zone):
+        self.gcp_facade = gcp_facade
+        self.project_id = project_id
+        self.zone = zone
+
+    async def fetch_all(self):
```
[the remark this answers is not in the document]
Remi05: Indeed 😐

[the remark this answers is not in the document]
Remi05: Though, since InstanceDisks is a bit different (the disks are already in the Instance, they are not fetched), it's simpler to call `fetch_all()` for each of the disks.

ScoutSuite/output/data/html/partials/gcp/left_menu_for_gcp_region.html
Remi05: Merged and took care of it!

PR #320, opened 2019-04-09
Enhancement/Additional AWS exception handling take 2

j4v:
Addressed comments made by @misg in https://github.com/nccgroup/ScoutSuite/pull/312.

I also changed the variable name for the `run_concurrently` and `map_concurrently` functions, is there a reason for using abreviations? They aren't reserved names.

### DIFF
--- a/ScoutSuite/providers/aws/facade/emr.py
+++ b/ScoutSuite/providers/aws/facade/emr.py
@@ -8,13 +8,13 @@ class EMRFacade(AWSBaseFacade):
     async def get_clusters(self, region):
         cluster_list = await AWSFacadeUtils.get_all_pages('emr', region, self.session, 'list_clusters', 'Clusters')
         cluster_ids = [cluster['Id'] for cluster in cluster_list]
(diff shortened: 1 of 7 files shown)


misg:
Looks good, thanks!
I took advantage of your PR to address https://github.com/nccgroup/ScoutSuite-Proprietary/pull/128/files#r273194877, see my last two commits. The problem in your current implementation was about handling exceptions occuring during executions of the `coroutine` passed to `map_concurrently`. The way you handled it (with an `except` that returns `None`, see https://github.com/nccgroup/ScoutSuite/pull/312/files#diff-556985dc1c1313b3cd49400717965473R32 for an example in SQS facade) would lead to a list returned by `map_concurrently` containing some `None` values that would have not been handled by the parsing process, leading to unexpected crashes.

The way I handled it is fairly simple, instead of returning `None` in case of exception in the `coroutine` passed to `map_concurrently`, I propagate the exception with `raise` so `map_concurrently` can catch it and ignore it to only store result of the executions of `coroutine` that have succeeded.

PR #323, opened 2019-04-09
bugfix/packaging

zer0x64:
This PR aims to fix packaging issues introduced since the last release. The most notable issue was that actual code was in the Scout.py file, which is not and cannot(as far as I know and in a clean way) be included in the package as it is not in the ScoutSuite/ folder.

### DIFF
--- a/Scout.py
+++ b/Scout.py
@@ -2,41 +2,7 @@
 
 import sys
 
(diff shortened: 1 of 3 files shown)


zer0x64:
Note that I also renamed the `Scout` script created by the package to `scout`, as this change is intended for the 5.0.0 release

PR #326, opened 2019-04-10
Refactor file hierarchy

j4v:
What this PR changes:
- Rename `Scout.py` to `scout.py` (PEP8)
- Move content of `ScoutSuite/scout.py` to `ScoutSuite/__main__.py`
- Move `main()` function from `Scout.py` to `ScoutSuite/__main__.py` and rename it as `run_from_cli()`
- Update `setup.py` to execute `__main__:run_from_cli`

The only thing I'm still a bit unhappy with is the uppercases in the module name (ScoutSuite) as it goes against PEP8 (https://www.python.org/dev/peps/pep-0008/#package-and-module-names), but changing it would be death.


j4v:
@zer0x64 if this looks good to you I'll fix the tests [and merge]

zer0x64:
Just need to fix the tests, but it looks good! I'll test it today

j4v:
@zer0x64 fixed :smile:

j4v:
> I just changed the command line tool from `Scout` to `scout`,

Great :smile: 

What do you have to put when running through pip? `ScoutSuite`, `scout`?

zer0x64:
I think it's still ScoutSuite? It would work like this:
```
pip install ScoutSuite
scout aws -f
```

j4v:
> I think it's still ScoutSuite? It would work like this:

Yep you're right, I've updated the README :smile:

PR #330, opened 2019-04-10
Removed duplicate requirement

vifor2:
Removed double requirement for `iampoliciesgonewild` which caused the following message and prevented from installing requirements when used on WSL (bash) :
image
And was added [here](https://github.com/nccgroup/ScoutSuite/commit/28e6d1f3e4bd4b068c0b0caf31eb718743919d3c#diff-b4ef698db8ca845e5845c4618278f29a)


vifor2:
> Would remove the one in Core instead of the one in AWS, since it only used in AWS

Done!

PR #344, opened 2019-04-11
Bugfix/peering connections revival

Aboisier:
Brought the peering connections back to life.


Aboisier:
> Looks good, but someone with AWS setup should also test it imo!
Sure, but why this particular PR haha

Remi05:
> > Looks good, but someone with AWS setup should also test it imo!
> 
> Sure, but why this particular PR haha

Not just this one in particular, I'm usually not the only reviewer so I don't mention it then! 😉

Aboisier:
> Not just this one in particular, I'm usually not the only one to have reviewed so I don't mention it then! 😉

Oh indeed, I usually wait for two reviews at least! I was wondering if there was something you though might not work in the PR. Thanks!

ScoutSuite/providers/aws/provider.py
```diff
@@ -428,6 +428,32 @@ def _match_instances_and_roles(self):
...
+
+        # Create a list of peering connection IDs in each VPC
+        info = 'AccepterVpcInfo' if current_config['AccepterVpcInfo']['OwnerId'] == self.account_id else 'RequesterVpcInfo'
+        region = current_path[current_path.index('regions') + 1]
+        vpc_id = current_config[info]['VpcId']
+        if vpc_id not in self.services['vpc']['regions'][region]['vpcs']:
+            region = current_config['AccepterVpcInfo']['Region']
+            target = self.services['vpc']['regions'][region]['vpcs'][vpc_id]
+        else:
+            target = self.services['vpc']['regions'][region]['vpcs'][vpc_id]
```
vifor2: Redudant assignation `target = self.services['vpc']['regions'][region]['vpcs'][vpc_id]`

ScoutSuite/providers/aws/facade/ec2.py
```diff
@@ -159,3 +161,7 @@ class EC2Facade(AWSBaseFacade):
         subnet['flow_logs'] = \
             [flow_log for flow_log in self.flow_logs_cache[region]
              if flow_log['ResourceId'] == subnet['SubnetId'] or flow_log['ResourceId'] == subnet['VpcId']]
+
+    async def get_peering_connections(self, region):
+        peering_connections = await AWSFacadeUtils.get_all_pages('ec2', region, self.session, 'describe_vpc_peering_connections', 'VpcPeeringConnections')
```
misg: Could you add some exception handling there in case of API failure?

PR #350, opened 2019-04-12
GCP-related cleanup

Remi05:
Cleaned up GCP resources, utils, provider, service config and more. Mainly removed dead/useless code and moved a few things.

Please let me know if there are other things that you think should also be cleaned for GCP!


ScoutSuite/providers/gcp/configs/base.py
```diff
@@ -1,230 +1,10 @@
...
-from google.api_core.exceptions import PermissionDenied
-from google.cloud import container_v1
-from googleapiclient.errors import HttpError
-
-from ScoutSuite.core.console import print_exception, print_error
-
 from ScoutSuite.providers.base.configs.base import BaseConfig
-from ScoutSuite.providers.gcp.utils import gcp_connect_service
 
 class GCPBaseConfig(BaseConfig):
```
Aboisier: Can't we just delete this file completely, since we aren't using configs anymore?

ScoutSuite/providers/gcp/facade/base.py
```diff
@@ -8,11 +8,28 @@ def __init__(self, client_name: str, client_version: str):
...
-        return self._build_client()
\ No newline at end of file
+        return self._build_client()
+
+
+class MemoryCache:
+    """
+    Workaround https://github.com/googleapis/google-api-python-client/issues/325#issuecomment-274349841
+    """
+    _CACHE = {}
```
Aboisier: Why is this capital? Is it some kind of python convention?

Remi05: I doubt it is, I simply moved the class from another location, it was already like this. I'll rename it!

ScoutSuite/providers/gcp/configs/base.py
Remi05: True

PR #354, opened 2019-04-13
Refactoring/183 cleanup

Aboisier:
* I moved the composite resources to the base class. I still kept the empty providers' classes (AWSCompositeResources, for example) so we don't have to go through each file if we want to add some provider specific logic. 

* I made the scopes more explicit. Before, AWS and Azure resources would receive a scope dictionary. Now they explicitly define the information they need in their constructor, and the scope dictionary is unpacked before it is passed to the class by the fetch children method.

* I renamed `gcp_facade` to `facade` so it was consistent with the other providers and to make it work with composite resource.

Don't worry about the 91 files, it's mostly small changes.

~~After that, I think we can remove **kwargs from `fetch_all`, and the cleanup is pretty much done.~~


Aboisier:
> Great stuff! Appears to be working just fine for GCP btw!

Noice, thanks for testing! A bit of Azure testing would be much appreciated, @misg!

misg:
Great work! :1st_place_medal: 

I just pushed a bunch of commits to finalize this cleanup:
* I modified a bit the composite resources to handle Azure services, they were broken.
* I removed all `**kwargs` and `credentials` parameters that were now unused and useless.
* I added a constructor to the base class `Resources` that takes a facade.
* I removed all empty `__init__` files.
* I factorized all concurrent parts of gcp with `_fetch_children_of_all_resources`.
* I fixed some bugs/broken services (Azure ad administrators were not handled properly in `sqldatabase` service, `ec2` was broken because of the use of `owner_id` in some scopes (I moved this to the facade, it's related to the credentials and thus shouldn't be in the resource classes), `emr` was broken as well and Azure private `appservice` was not imported as expected)
* And finally, I fixed plenty of PEP8 issues.

Oh and the [proprietary part](https://github.com/nccgroup/ScoutSuite-Proprietary/pull/132) of this PR is now available, please take a look at it.

Aboisier:
Looks really good, thanks for the help! Tested on AWS, works as expected. 🎉

Issue #298, opened 2019-04-03
Add support back for AWS ELB policies, VPC peering connections and hosted zones


misg:
Some resources were removed during refactoring #183 but they were displayed in Scout Suite reports and thus potentially useful (and used). They should be back.

- [x] ~~`peering_connections`, removed in #273.~~ (**Brought back in #344.**)
- [x] `elb_policies` resources, removed in #257. (**In progress, #358.**)
- [x] `hosted_zones` resources, removed in #267.

j4v:
Implemented in `develop`.

j4v:
Closing as this is implemented in `master`.

PR #358, opened 2019-04-15
Bugfix/elb policies revival

misg:
This PR aims to bring back elb policies to life, as part of #298.
Although, I didn't manage to really test it (I tested it against our aws test account, with a load balancer set up but I didn't find a way for adding a name to my elb policy (which is not retrieved in the end for that reason)).
Could you take a look, @j4v ?


j4v:
@misg I think all you where missing was an HTTPS listener. I've configured one so you should now get results.

misg:
Oh indeed, nice catch! I had a bug although, I just fixed it. I'm able to retrieve elb policies now, their names are correctly displayed in Scout Suite reports. But the fields `Protocols`, `Options` and `Ciphers` are empty, not sure if it's expected or not (regarding the elb policy I use for testing), there could be some issue in the parsing/displaying.

j4v:
@misg this should now be fixed, there was a minor bug in bringing back Scout2 code.

PR #362, opened 2019-04-15
Tested refactoring, file path handling modifications & minor bug fixes

j4v:
This PR implements the following changes:
- Refactors the way file names / paths were handled, as the previous implementation was a bit confusing.
  - Removed some AWS legacy code from the /output/ folder (a lot of mentions of `profile`
- Tweaked report naming for AWS/GCP/Azure
- Removed `coding: utf-8` statements
- Fixed a few bugs/error handling in GCP
- Added a `--version` CLI argument
- Removed Azure from beta in README
- Renamed what was probably a copy-pasta mistake

Tested:
- AWS, GCP, Azure providers
- `--local`, `--update`, `--ruleset`, `--exceptions`, `--report-name`, `--repor-dir` flags

Issues:
- ~~Blank page when accessing server report @zer0x64~~
- ~~Exception `services.py L47: Could not fetch vpc configuration: 'flow_logs_count'` when running against AWS~~ @misg @Aboisier 
- For GCP it looks like the project/folder/organization filtering is no longer there? @Remi05 
  - Looks like all the projects are always being scanned


j4v:
> Let me know if it's still happening,

Fixed :heart_decoration:

vifor2:
Fixed the issue when viewing a report for SQLite, the import used the wrong path.

@j4v 
> Blank page when accessing server report @zer0x64

It is blank when viewing 127.0.0.1:8000 since there aren't any routes currently allocated to that address. If you would like to view the data from the server you must use a valid route such as : `http://127.0.0.1:8000/api/data?key=` to view only the elements in the first layer or `http://127.0.0.1:8000/api/full` to view everything. Which reminds me, I'll go write on that subject in the wiki.

PR #376, opened 2019-04-25
Add support for AWS config resources

gebailey:
This change adds AWS config resources to the collected data, and defines one new finding to verify if there are any configuration recorders for each enabled region.

The finding is disabled in the default and detailed rule sets.


j4v:
Closing as tracked in https://github.com/nccgroup/ScoutSuite/pull/377.

Issue #407, opened 2019-05-13
Incorrect CMD when not running through CLI


thommor:
When running ScoutSuite through another python program (as an import) the `cmd` in `last_run` is the command of the overarching program, not what would of been ran if using ScoutSuite from the CLI.

It would be useful that if ScoutSuite is ran as an import that the `cmd` is set to what would have been run if running from the CLI.

j4v:
Would it work better if instead of the CLI value (taken from `sys.argv`) it was a dict of the non-`None` values passed to `_run()`?

In the report, instead of displaying "Command", we could display these values.

thommor:
@j4v Yeah that works for me

j4v:
@thommor looks something like this:

![sc_2019-05-17_16h09m05s](https://user-images.githubusercontent.com/4206926/57933713-4481fe00-78be-11e9-8e2e-e604ca2fbdda.jpg)

I don't think it looks as "nice" as before especially since the parameter names are the ones in the code not the CLI but it does the job.

Comments/Suggestions?

j4v:
Closing as merged into `develop`.

PR #414, opened 2019-05-17
Update execution details to work when Scout is run programatically

j4v:
Implements https://github.com/nccgroup/ScoutSuite/issues/407.


j4v:
Updated the implementation to only include desired fields:

sc_2019-05-17_18h26m39s

Will only print services / skipped services / regions if they are included.

thommor:
Looks good to me!

Issue #449, opened 2019-06-17
When using profiles to role switch to other accounts Scout scans the default profile


FishermansEnemy:
I have a default profile with AWS ID and Secret Key set in credentials (deafult). I have two profiles in config, profilea and profileb which I role switch into to get access to their accounts. AWS cli works fine and when using these profiles from AWS cli I can see the three different accounts.

When I run scout with the --profile profilea or --profile profileb it runs and produces the same report for the default profile, along with the same account number.

j4v:
I think this is the same as https://github.com/nccgroup/ScoutSuite/issues/425. 

Can you show what you mean by "AWS cli works fine"? We're using boto3's implementation for authentication (https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html) so in theory it should work.

FishermansEnemy:
That does look like the same issue as I am having. By “AWS Cli works fine” I mean I can issue commands as one of my two profiles and see contents from the accounts that each profile have a role in.

j4v:
Thanks @FishermansEnemy, will have to look into how boto3 handles this as it might not be as straightforward as what their doc says.
Using the config file and "source_profile" seems to be part of the issue.

j4v:
@FishermansEnemy this should now be resolved in branch https://github.com/nccgroup/ScoutSuite/tree/bugfix/aws-session-handling - could you please test?

j4v:
Closing, this should be resolved by https://github.com/nccgroup/ScoutSuite/pull/460. If not please reopen the issue.

Issue #425, opened 2019-05-21
Receiving bucket list gives AccessError even though it works with aws_cli


hashier:
`I'm using,`

macOS 10.14.5 
Python 3.7.3

with ScoutSuite installed via pip3:
$ pip3 freeze | grep -i scout
ScoutSuite==5.0.post3

and I tried as well the latest git checkout:
$ git rev-parse HEAD
5a498fbd9fc558d63409123ab5aba19ad4e5097d

`What did you do?`

When I call:
```
$ aws s3 ls --profile security
Enter MFA code for arn:aws:iam::xxxx:mfa/username:
2018-04-12 16:20:13 foo
2018-04-20 14:43:42 bar
[...]
```

I can see all s3 buckets, but when I try fetch the buckets via Scout I end up with the following error:

```
$ python3 scout.py aws --profile security --services s3 --no-browser
2019-05-21 16:42:21 PinkRobin.localdomain scout[33071] INFO Launching Scout
2019-05-21 16:42:21 PinkRobin.localdomain scout[33071] INFO Authenticating to cloud provider
Enter MFA code for arn:aws:iam::xxxxx:mfa/username:
2019-05-21 16:42:29 PinkRobin.localdomain scout[33071] INFO Gathering data from APIs
2019-05-21 16:42:29 PinkRobin.localdomain scout[33071] INFO Fetching resources for the S3 service
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] ERROR s3.py L15: Failed to list buckets: An error occurred (AccessDenied) when calling the ListBuckets operation: Access Denied
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] INFO Running rule engine
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] INFO Applying display filters
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] INFO Saving data to scoutsuite-report/scoutsuite-results/scoutsuite_results_aws-security.js
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] INFO Saving data to scoutsuite-report/scoutsuite-results/scoutsuite_exceptions_aws-security.js
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] INFO Saving data to scoutsuite-report/scoutsuite-results/scoutsuite_errors_aws-security.json
2019-05-21 16:42:30 PinkRobin.localdomain scout[33071] INFO Creating scoutsuite-report/aws-security.html
[...]

PR #460, opened 2019-07-01
Fix role switching bug

j4v:
Should fix:
- https://github.com/nccgroup/ScoutSuite/issues/449
- https://github.com/nccgroup/ScoutSuite/issues/425

### DIFF
--- a/ScoutSuite/providers/aws/authentication_strategy.py
+++ b/ScoutSuite/providers/aws/authentication_strategy.py
@@ -1,6 +1,13 @@
 import boto3
 
 from ScoutSuite.providers.base.authentication_strategy import AuthenticationStrategy, AuthenticationException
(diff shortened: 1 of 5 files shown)


j4v:
> Fixed the issue

Thanks @AlfredLebeau, did you test in a MFA scenario or just normal role assumption?

AlfredLebeau:
Normal role assumption

estemendoza:
Hi guys. Thanks for fixing this.

I just wanted to say that this broke our current scout scans for AWS and we couldn't make it work until I found this fix on the develop branch because we also tried on master and it was still failing.

Is this not happening to everybody using the latest release?

j4v:
> Is this not happening to everybody using the latest release?

The latest release / `master` doesn't include this PR?

Issue #91, opened 2019-01-03
Feature Request - handle API rate limiting


museadmin:
Hi

AWS has a weakness in that it rate limits API calls in a blanket fashion. So if I run Scout Suite against our account while the platform is busy, it can break production.

I'm specifically hitting issues with autoscaling failing because AWS draws no distinction between the scanner making API calls on the CLI and their own autoscaling feature making API calls. Because the latter then fails to bring up the instance, the environment ends up in an indeterminate mess and the platform team have to scramble.

It would help if you had an option to run in single threaded mode to limit the amount of API calls being made.  Perhaps go even further and add an ability to rate limit your calls from within the application, so that AWS didn't kick in and do so itself.

j4v:
Adding single threaded mode should be a quick fix.

Scout does handle rate limiting, but that won't stop the situation you describe from happening.

j4v:
The https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/base/configs/threads.py file configures the amount of threads created when using the `--thread-config` option (1 to 5, default at 4). As you can see, setting `--thread-config 1` will spin 2 threads, 1 for listing resources and another one for parsing them (parsing typically does not interact with the API). 

Have you tried using this option to limit impact on API usage?

museadmin:
Thanks for the feedback Xavier, 

No I wasn't aware of that and will certainly make use of it. I only run the suite once every two weeks (a sprint) so will try it next time.

When you say that the suite handles rate limiting...

Most people handle this by implementing an incremental backoff, which if that is the case here, would not help as the problem is really within the way they are implementing autoscaling. 

I guess the real answer is persuading the platform guys to implement something to handle the failure. The fact that they struggle to fix the broken instances suggests to me that they could do better with the way the system is architected.

Once again, thanks for the thread advice.

Brad

museadmin:
Thanks for looking at this guys. Can I add these comments to this ticket because I think there is a case for going further with a remediation:

E.g. I see this error even when running with --max-workers 1 

2019-06-27 09:12:18 MAC2802.local scout[5467] ERROR ec2.py L127: Failed to describe EC2 snapshot attributes: An error occurred (RequestLimitExceeded) when calling the DescribeSnapshotAttribute operation (reached max retries: 4): Request limit exceeded.

AWS have blighted pretty much all of the scanning tools I’ve evaluated, yours being one of the best, because of a number of reasons:

[...]

PR #461, opened 2019-07-02
Experimental/handle AWS rate limiting

j4v:
Partial fix for https://github.com/nccgroup/ScoutSuite/issues/91, implements backoff & retry mechanism.

TODOs:
- ~~Create a function per provider to assess if rate limiting is happening~~ to be done once we hit any issues with rate limiting in other providers
- ~~Also work towards minimizing API calls in AWS provider~~ this is mostly the case, instead will implement request rate limiting
  - For some services (e.g. IAM), API calls that match resources (e.g. users and groups) could be done locally in the `finalize` method instead of in the facades through `get_and_set` methods.


j4v:
> The is_throttled function could be removed? It doesn't appear to be called anywhere

I've moved the check back to AWS to keep things a bit cleaner. If we hit throttling with another provider we can implement a similar method.

PR #472, opened 2019-07-19
Include tags in S3 bucket data

gebailey:
Retrieves tags associated with S3 buckets.  This would allow custom rules to be
written that could use these tags when deciding whether or not to generate a
finding.


j4v:
Great addition @gebailey, tags are indeed a very nice feature to base rules on.

Issue #471, opened 2019-07-19
Use of "N/A" in AWS EC2 SGs is confusing


j4v:
Instead should use either `*` or `1-65535`:

![sc_2019-07-19_16h42m19s](https://user-images.githubusercontent.com/4206926/61543808-78a3a780-aa44-11e9-9666-fc127aa51036.jpg)

j4v:
Fixed in `develop`.

PR #476, opened 2019-07-22
Report showing N/A for all ports was confusing so changed to 1-65535

3vilsaint:


j4v:
Fixes https://github.com/nccgroup/ScoutSuite/issues/471.

PR #479, opened 2019-07-24
Enhancement/Azure GraphRBAC

j4v:
Adds support for Azure GraphRBAC - WIP.

~~Current issue is with the resource endpoint hardcoded into https://github.com/nccgroup/ScoutSuite/blob/6ba3ba7d7ce71d83750fde3a0e6ce767c7355180/ScoutSuite/providers/azure/facade/graphrbac.py, which breaks when scanning other services.~~


j4v:
@JJmako I've fixed authentication so it now works for GraphRBAC & other services. Only thing I haven't been able to test is MSI authentication (would need to run from a VM).

Ready to add support for users/groups/etc!

Issue #473, opened 2019-07-21
Route53 hosted zones


benghaun:
The `hosted_zones` resource for route53 was removed in #267, and a case was made to bring it back in #298. However, #298 was closed although the resource has not been added back. Aside from the fact that `hosted_zones` are viewable in the ScoutSuite report, having information about all DNS records in Route53 has also been helpful for further security checks.

Are there plans to add the `hosted_zones` resource back? If no one is working on it, I will be happy to take it up. 

Thank you!

j4v:
Looks like there was some confusion in the issue(s)/PR(s) and indeed they weren't brought back. No objection to doing so, we just haven't gotten around to doing it.

@benghaun if you can submit a PR I'll be happy to review/merge.

j4v:
Closing as implemented in https://github.com/nccgroup/ScoutSuite/pull/493.

PR #482, opened 2019-07-27
Added hosted zones and removed regions from route53

benghaun:
As per #473, this PR adds the hosted zones resource back into ScoutSuite for Route53.

While I was doing this, i also realised that the `Route53` config class was inheriting from `ScoutSuite.providers.aws.resources.regions.Regions`, which didn't make sense since Route53 is a service that has no region selection. However I also realised that for domains specifically, there is no global endpoint but only one regional endpoint, which is `us-east-1`, which was probably the reason why the domains were placed under `us-east-1` Nonetheless, it is essentially a global service and it makes more sense to place it as a global service. 

With this PR, the `Route53` config class was changed to inherit from `AWSCompositeResources` instead. This means that both the domains and the newly added hosted zones will not be classified as resources under the `us-east-1` region.

If I recall correctly, domains were also previously classified as global services. Do let me know if there are any problems with classifying them as a global service. Hosted zones are most definitely global as the api endpoint for it is also global.

### DIFF
--- a/ScoutSuite/providers/aws/facade/route53.py
+++ b/ScoutSuite/providers/aws/facade/route53.py
@@ -4,9 +4,21 @@ from ScoutSuite.providers.aws.facade.utils import AWSFacadeUtils
 
 
 class Route53Facade(AWSBaseFacade):
-    async def get_domains(self, region):
+    async def get_domains(self):
         try:
-            return await AWSFacadeUtils.get_all_pages('route53domains', region, self.session, 'list_domains', 'Domains')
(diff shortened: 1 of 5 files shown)


j4v:
Thanks for this, will review asap (I'm currently on leave).

j4v:
@benghaun apologies for the delay, just got back from leave.

I've reviewed this and made some changes, please review https://github.com/nccgroup/ScoutSuite/pull/493. 

I reverted the resource to a regional one as in the future AWS might add another endpoint for this. I've also improved the report and added a partial for hosted zones.

PR #498, opened 2019-08-16
Enhancement/Azure Users & Groups

JJmako:
Added groups in Graphrbac.
Added "Guest users" detection rule.


j4v:
@JJmako I've made a few changes and added a partial to display groups in the report.

What's missing is to cross-reference users and groups (so that when viewing a user you can see to what groups he belongs, and when viewing a group you can see its users). Refer to an AWS report to see what is should look like.

ScoutSuite/providers/azure/rules/findings/graphrbac-guest-users.json
```diff
@@ -0,0 +1,11 @@
+{
+    "description": "Guest Users in Use",
+	"rationale": "<b>Description:</b><br><br>Avoid creating guest users, as they are typically added outside your employee on-boarding/off-boarding process and could potentially be overlooked indefinitely leading to a potential vulnerability.<br><br><b>References:</b><ul><li>CIS Microsoft Azure Foundations v1.1.0 1.3</li></ul>",
+    "dashboard_name": "Users",
+    "path": "graphrbac.users.id",
+    "display_path": "graphrbac.users.id",
+    "conditions": [ "and",
+        [ "graphrbac.users.id.user_type", "equal", "Guest" ]
+    ],
+    "id_suffix": "user_type"
```
j4v: This should be set to the last suffix in the HTML object's ID so that it can be highlighted.

j4v: Which in this case is `user_type` (I changed it from the value you set).

j4v:
Reopening as there are some unmerged commits that are not also found in https://github.com/nccgroup/ScoutSuite/pull/505/commits.

j4v:
Closing, code is now in #539.

PR #517, opened 2019-08-30
feature/516-replace-iampoliciesgonewild-dependency

### LINKED ISSUE #516: Replace iampoliciesgonewild dependency
ScoutSuite still uses the module `iampoliciesgonewild` which was put together by Patrick Kelley (previously @monkeysecurity but changed his github handle to @scriptsrc at some point) who works for Netflix. The `iampoliciesgonewild` module was merged into Netflix's [policyuniverse](https://github.com/Netflix-Skunkworks/policyuniverse) package and can be found at https://github.com/Netflix-Skunkworks/policyuniverse/blob/master/policyuniverse/expander_minimizer.py which is a drop in replacement for `iampoliciesgonewild`.

Two reasons for updating are:
1. `iampoliciesgonewild` is no longer maintained.
2. An org that wants to use ScoutSuite but also reviews dependencies will approve ScoutSuite quicker if it uses more reputable and maintained dependencies.

I will have a followup PR to replace the module with `policyuniverse`.

jeffxf:
- Replaces import in core/conditions.py
- Replaces dependency in requirements.txt

Closes #516

### DIFF
--- a/ScoutSuite/core/conditions.py
+++ b/ScoutSuite/core/conditions.py
@@ -4,7 +4,7 @@ import json
 import netaddr
 import re
 
-from iampoliciesgonewild import get_actions_from_statement, _expand_wildcard_action
+from policyuniverse.expander_minimizer import get_actions_from_statement, _expand_wildcard_action
 
 from ScoutSuite.core.console import print_error, print_exception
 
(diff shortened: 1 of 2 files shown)


j4v:
thanks @jeffxf, will review ASAP.

PR #532, opened 2019-09-23
Fixed scoping issue in _get_and_set_inline_policies

wstrei:
The current implementation scopes variables passed to the lambda function such that an entity with multiple inline policies will have get_policy_method called with the same policy_name for each policy. Creating a policy_name parameter with a default fixes this issue.


j4v:
Fix pushed to `develop` as per https://github.com/nccgroup/ScoutSuite/commit/d4751c2a8f06f5efef03d76e6c3ab52f4684c7b1.

PR #534, opened 2019-09-24
Retrieve EC2 security group tags

gebailey:
Retrieves tags associated with EC2 security groups.  This would allow custom rules to be written that could use these tags when deciding whether or not to generate a finding.


gebailey:
Sorry this was based off of master vs. develop; if you need/want me to create a new pull request against the develop branch, let me know.  Thanks!

gebailey:
closing; I will resubmit against the develop branch

PR #543, opened 2019-10-10
Enhancement/Added descriptions to IAM roles

Lucasargh:
When reviewing ScoutSuite reports, the additional context provided by descriptions is extremely valuable. This PR adds descriptions to IAM roles specifically.


j4v:
Thanks, made some changes and merged to `develop`.

PR #545, opened 2019-10-10
Add check, rule and output covering RDS Certificate Authority deprecation in AWS environments

scianos:


j4v:
Many thanks for this @scianos, accepted and merged :100: .

Issue #551, opened 2019-10-14
Docker support


kreatemore:
Hi!

I wanted to experiment with ScoutSuite, and I have created a Dockerfile so I can run it without virtualenv - is this something you might be interested in?

I'll create a PR shortly, and then you can decide if we should pursue this.

On another note, I must say ScoutSuite is quite nice, I'm quite pleased with the generated report!

j4v:
Happy you've found the tool useful @kreatemore.

I don't really see the benefit of leveraging docker to run ScoutSuite, considering it's pretty trivial with virtualenv/pip. Maybe I'm missing something / specific use cases?

kreatemore:
Thanks for the quick reply - I think it might be a highly subjective use-case, I personally prefer working in Docker, rather than having packages installed on my host.

Apart from this - and I'm still exploring this - it might be useful if you want to run this as part of an automated workflow.

The PR stands ready, if you'd like to see it, I'll submit - but if you don't feel the need for this, feel free to close this issue, and I won't create a PR

thommor:
I don’t think it would hurt to have one.

The maintenance and upkeep of a Dockerfile is minimal.

j4v:
> I don’t think it would hurt to have one.

Ok then @kreatemore feel free to submit a PR :smile: 

> I personally prefer working in Docker, rather than having packages installed on my host.

That's why you use virtual environments!!!

j4v:
Thanks for the contribution @kreatemore, this is now merged!

PR #552, opened 2019-10-14
#551 Adds Docker support

kreatemore:
👋 this is the proposed Dockerfile - unfortunately couldn't use an alpine based image, this is the second smallest I could find.

Not sure if I can include changes for the wiki, but this should be the way of installing & running ScoutSuite with Docker:

## Install
0. clone the repository, and cd to the dir
1. `docker build -t scoutsuite:latest .`

## Run
```
docker run --rm -t \
-v /Users/me/.aws:/root/.aws:ro \
-v "$(pwd)/results:/opt/scoutsuite-report" \
scoutsuite:latest \
aws
```

If you'd like to add further clarification to these commands, it will:
* run the image we just built with the `tty` flag, and the container will be removed once it finishes
* will create a read-only volume to `/Users/me/.aws` (where `/Users/me` should be changed to the home directory of whoever runs it) so the aws creds are available from within the container. 
* will create another volume to $PWD/results so the generated report is accessible from the host
* finally (after `scoutsuite:latest`) you just pass in arguments as you normally would to the script itself

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:slim
+
+COPY . /opt
+
+WORKDIR /opt
+RUN pip install -r requirements.txt
+RUN mkdir /root/.aws
+
+ENTRYPOINT ["python", "scout.py"]


j4v:
@thommor could you have a look please?

j4v:
> @j4v One of the tests is failing but I cant see this being caused by the Dockerfile, can you confirm?

re-ran the tests and it went fine - for some reason builds sometime fail...

> The wiki will need updating

will do now

PR #553, opened 2019-10-16
Added Subnet NSG

JJmako:
Improve support for Azure Security Groups.


ScoutSuite/providers/azure/resources/network/network_interfaces.py
```diff
@@ -62,5 +62,7 @@ def _parse_network_interface(self, raw_network_interface):
         network_interface_dict['ip_configuration']['name'] = ip_configuration.name
         network_interface_dict['ip_configuration']['etag'] = ip_configuration.etag
 
+        network_interface_dict['ip_configuration']['subnet_security_group'] = ip_configuration.subnet.network_security_group
```
j4v: This seems to be `None` even if there is a NSG configured at the subnet level?

PR #560, opened 2019-10-31
Include tags in ELB and ELBv2 data

gebailey:
Retrieves tags associated with ELB and ELBv2 load balancers.  This would allow
custom rules to be written that could use these tags when deciding whether or
not to generate a finding.


j4v:
Thanks for this @gebailey. Could you include this in the report generation too? If you're going to flag resources based on tags, ideally there should be a way of visually confirming findings.

gebailey:
@j4v I added the tags to the reports.  I can confirm that the ELB report optionally shows the tags, if they're present.  The ELBv2 report doesn't load for me (even before I made any of the changes), so I'm unable to confirm whether it's working correctly, but the change should be the same as for ELB.

PR #562, opened 2019-11-04
adds the ability to create binary packages

northdpole:
ScoutSuite needs python and pip to run which makes running it in Dockerfiles a pain as you need to bake python and other things in your container.

Having the ability to create a .pex file which can be executed as a binary allows for running ScoutSuite in minimal containers.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -54,3 +54,6 @@ aliyun-python-sdk-kms>=2.6.0
 
 # Oracle Cloud Infrastructure Provider
 oci>=2.2.4
+
+# packaging
+pex==1.6.12
--- a/tools/gen-package.sh
+++ b/tools/gen-package.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+pex scoutsuite --compile -v -c scout -o scout.pex


northdpole:
> Looks good to me although i'm unsure if the binary should be included in the repo
> 
> @j4v thoughts?
in my point of view, makes it more legit, I can download the binary from the release and compare it to the one in the repo or generate the binary and make sure they match.
Happy to remove it if not required :-)

thommor:
> > Looks good to me although i'm unsure if the binary should be included in the repo
> > @j4v thoughts?
> 
> in my point of view, makes it more legit, I can download the binary from the release and compare it to the one in the repo or generate the binary and make sure they match.
> Happy to remove it if not required :-)

@j4v has the final say but in my opinion I'd say include it in each release as a binary download (which we can build for each release) but would avoid having it live in the repo permanently as it risks being forgotten and not updated when new code is added

j4v:
If this is a one-liner can't you just integrate it in your build process? 

> it risks being forgotten and not updated when new code is added

This is my main concern.

northdpole:
> If this is a one-liner can't you just integrate it in your build process?
> 
> > it risks being forgotten and not updated when new code is added
> 
> This is my main concern.

I can, and happy to do so if having binary releases isn't something you're interested in.

VJftw:
Maybe generating the binary and adding it as an asset when you create GitHub Releases is better as it'll be downloadable from here: https://github.com/nccgroup/ScoutSuite/releases and not committed to the code-base.

j4v:
Closing, see comments in https://github.com/nccgroup/ScoutSuite/issues/563.

Issue #554, opened 2019-10-17
FR: Dynamically pull AWS IP ranges


aph3rson:
There are past issues (e.g. #335 and #336) where the list of AWS IP ranges is called out as out-of-date. At the moment, it hasn't been updated since April.

AWS publishes their most up-to-date IP ranges at https://ip-ranges.amazonaws.com/ip-ranges.json. It might be easier to load the file contents from their website, rather than loading it from a static copy in the repo. Some caching logic might be good here, at least for the duration of the scan, but otherwise it should be minimally impactful on scan duration.

j4v:
Thanks, putting this on the stack - feel free to submit a PR :smile:.

klauern:
So it looks like it's here, which seems pretty deeply nested within a lot of other nesting, so I'm thinking that there will need to be a lot more tests to support it.

https://github.com/nccgroup/ScoutSuite/blob/6cadac7096cb80e5b26d66b219c3e464c9a4faa5/ScoutSuite/core/rule.py#L107-L123

j4v:
@klauern a CI job to pull/update the file would work, just have to get down to implementing it.

j4v:
Closing, this has been included in the release process.

PR #566, opened 2019-11-06
Update AWS IP Address ranges

klauern:
This changes regularly, so if this is being used, it should be updated frequently.  I added an updated `aws.json` file, as well as a small script to re-download and store it.  Hopefully this can be integrated into some kind of CI/CD automation tool so this won't go out-of-date as often.


j4v:
Thanks @klauern, merged. 

This should indeed be integrated into Travis CI (as mentioned here https://github.com/nccgroup/ScoutSuite/issues/554), but we haven't had the time to work on it. Happy for you to submit a PR on this too :smile: .

PR #583, opened 2019-11-26
Fixed rules for root IAM stuff, also updated test data set

BastienFaure:
The JSON path used to match stuff about the IAM root account has probably changed. All my recent scout runs had greyed out items for IAM root related findings.
The rules tried to access the data at path:
```
iam.credential_reports.<root_account>
```
The key `<root_account>` does not longer exists as there is a dictionary instead, with one entry per account.

I updated all related rules and pushed a new dataset for running the tests.


j4v:
Thanks @BastienFaure, will review/merge ASAP.

BastienFaure:
@j4v I got this message when running ScoutSuite but I can't tell if this is just a warning or an error. Thanks for receiving the PR :)
details

j4v:
> @j4v I got this message when running ScoutSuite but I can't tell if this is just a warning or an error. Thanks for receiving the PR :)
> details

That looks like the path to a few rules is broken [by the PR?]. `--debug` should provide more info.

bishopfaure:
Found and fixed the issue, I had one of the rules not being properly written.

j4v:
:100:

Issue #378, opened 2019-04-25
Left column menu disappearing


j4v:
Left column menu disappear when navigating a bit:

![sc_2019-04-25_18h44m52s](https://user-images.githubusercontent.com/4206926/56753202-60e0be00-678a-11e9-9988-428b84da5e8c.jpg)

Reloading makes it reappear.

j4v:
Potentially the same issue - when viewing resources affected by a finding for a number of services (e.g. SES), the left menu isn't shown. When viewing all the resources for the same service they are.

j4v:
Closing, this is now fixed in `develop`.

PR #587, opened 2019-12-10
Fix/Multiple UI Issues

### LINKED ISSUE #584: Missing information when revisiting elements of a report
When first visiting certain elements in the UI information will be populated but when re visiting them after navigating information is missing. Any other section in the UI that uses the same component seems to be blank after this as well. The component in question is the one that is used on the external attack surface pages.

Please see images below for an example.

Thanks

populated
missing

thommor:
Fixed some login in ScoutSuite.js to stop the UI thinking it needed to load render the data several times.

Fixed two rules that were breaking the UI as they don't display anything.

This will fix #584 and #378


j4v:
@thommor to confirm, both https://github.com/nccgroup/ScoutSuite/issues/584 & https://github.com/nccgroup/ScoutSuite/issues/378 are still TODOs, yes?

thommor:
@j4v yes these are still TODOs. Will look in to them more next week

PR #594, opened 2019-12-23
Enhancement/Improved test coverage

JJmako:
Improved coverage from 61% to 67%


JJmako:
There are some problems with CI that doesn't exists on local. I will check the errors and commit again without that lines.

PR #610, opened 2020-01-20
Add new SSL/TLS security policies 

trevorrea:
The test Older SSL/TLS policy generates a false positive as it does not include the newer policies as listed at https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies

Specifically:-

- ELBSecurityPolicy-FS-1-1-2019-08
- ELBSecurityPolicy-FS-1-2-2019-08
- ELBSecurityPolicy-FS-1-2-Res-2019-08

This MR adds the new policies to the list


j4v:
Thanks @trevorrea, looks good.

PR #613, opened 2020-01-22
Add Tags for RDS Database Instances

klauern:
We use tagging for a lot of auditing purposes, so having this data returned to us in a ScoutSuite run is very useful.  This PR adds those tags in.

One unfortunate side-effect is that this requires a new API call to [`ListTagsForResources`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ListTagsForResource.html)


klauern:
I'm not sure how I should test this addition.  Can someone point me to an example test that exists for RDS currently that I could add this to?

ScoutSuite/__main__.py
```diff
@@ -113,6 +113,8 @@ def run(provider,
     """
 
     loop = asyncio.get_event_loop()
+    if loop.is_closed():
```
j4v: Why are you adding this?

klauern: gah, this is unrelated.  This is so it will run in a Lambda.

klauern: I removed this.  I'll try to put it in another PR later

klauern:
One note I forgot to mention was that this page will have to add the `rds:ListTagsForResource` permission: https://github.com/nccgroup/ScoutSuite/wiki/AWS-Minimal-Privileges-Policy

j4v:
Looks :100: , merging.

> One note I forgot to mention was that this page will have to add the rds:ListTagsForResource permission: https://github.com/nccgroup/ScoutSuite/wiki/AWS-Minimal-Privileges-Policy

Thanks for mentioning :medal_military:

PR #614, opened 2020-01-23
feat: create event loop when existing one is closed

klauern:
This is to fix an issue with AWS Lambdas not having an event loop at the
start of the run.  If there is none present, this just creates one.


j4v:
Thanks, will test/merge ASAP. Also please make PRs against `develop` branch (see https://github.com/nccgroup/ScoutSuite/blob/master/CONTRIBUTING.md).

klauern:
oh, I'm sorry about that.  I will make note of that for future PR's.

PR #626, opened 2020-01-28
feat: add EC2 instance tags

klauern:
Adding EC2 Instance Tags

### DIFF
--- a/ScoutSuite/providers/aws/resources/ec2/instances.py
+++ b/ScoutSuite/providers/aws/resources/ec2/instances.py
@@ -28,7 +28,7 @@ class EC2Instances(AWSResources):
 
         get_name(raw_instance, instance, 'InstanceId')
         get_keys(raw_instance, instance,
-                 ['KeyName', 'LaunchTime', 'InstanceType', 'State', 'IamInstanceProfile', 'SubnetId'])
+                 ['KeyName', 'LaunchTime', 'InstanceType', 'State', 'IamInstanceProfile', 'SubnetId', 'TagSet'])
 
         instance['network_interfaces'] = {}
         for eni in raw_instance['NetworkInterfaces']:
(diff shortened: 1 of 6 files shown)


j4v:
@klauern is this ready for review/merge?

klauern:
I had to add a couple other EC2-related items (volumes, snapshots) to have tags in them as well.  This should be good to review at this time.

j4v:
@klauern looking at the resulting JSON, I do see the tags in the volumes and snapshots but not for the instances. Is this expected?

klauern:
Hm, I thought so, but maybe not.  I have it on my side, so it's probably on another branch that I need to pull in.

ScoutSuite/providers/aws/resources/ec2/instances.py
```diff
@@ -28,14 +28,15 @@ def __init__(self, facade: AWSFacade, region: str, vpc: str):
 
         get_name(raw_instance, instance, 'InstanceId')
         get_keys(raw_instance, instance,
-                 ['KeyName', 'LaunchTime', 'InstanceType', 'State', 'IamInstanceProfile', 'SubnetId'])
+                 ['KeyName', 'LaunchTime', 'InstanceType', 'State', 'IamInstanceProfile', 'SubnetId', 'TagSet'])
```
klauern: I added this field to include the TagSet.  I probably didn't include the flat-mapping of TagSet to Tags mapping, but I can do that.  We have a downstream ETL process to do some light massaging of data, so that's probably where I put it.

j4v: Sure, all I know is that I added tags to an EC2 instance but they didn't show up in the report. If you're happy with the implementation I have no issue with merging.

klauern: I'm having a lot of issues with the case-sensitivity of the branches, so I'm deleting this one and re-creating a PR witht he `Enhancements/` folder..

PR #632, opened 2020-02-03
Enhancement: add EC2 instance tags

klauern:
feat: add EC2 instance tags


j4v:
I've re-reviewed this and what I don't get is why it's required. Aren't tags already included for these 4 resources (under the `Tags` field)? It seems redundant to have the same values under both `Tags` and `tags`.

klauern:
I'm probably confused on how the architecture works.  I had thought that
you needed to extract the "Tags" field first, which is why it is referenced
in the `_parse_X` calls.  Then, I saw it used to pull out fields in the
`instance` or `volume`, and thought that from the internal representation,
everything that was in those resource types had to be mapped out.

-Nick Klauer

klauern:
Ah, one other thing I forgot (sorry, still waking up) is that the `_get_and_set_*` method call flattens the Tags field, since it's normally:

```json
"Tags": [
{
  "Key":   "key",
  "Value": "value"
}, ...
```

instead of 

```json
"Tags": [
{ "key": "value" }, ...
]
```

j4v:
> I had thought that
you needed to extract the "Tags" field first, which is why it is referenced
in the `_parse_X` calls.

Generally speaking you're correct. Our convention (which is also implemented by the handy https://github.com/nccgroup/ScoutSuite/wiki/Tools#process_raw_responsepy script) is that the parsing method would create a **new** dictionary and store all the relevant information. For a number of AWS services/resources (such as EC2), the implementation is a bit more lazy (legacy, legacy...) and instead of creating a new dict, the code modifies the raw object returned by boto3 (the AWS python library). That's why the tags are already there without doing anything.

Quick tip: if you run with the `--debug` CLI parameter, the resulting JSON report is prettified, which makes it easier to see what's being stored on disk.

PR #639, opened 2020-02-07
Enhancement/Add KMS resource data

klauern:
This pulls down the KMS information into ScoutSuite:

- Aliases
- Keys
  - KeyPolicies
  - Grants


j4v:
Merged with a few changes:
- Put aliases and policies into the keys
- Added an HTML partial to display resources

PR #643, opened 2020-02-09
Enhancement/Add black formatting

klauern:
There's no apparent pick for a formatter, so might just pick one.  Black
is 'good enough', but as long as there's consistency in PR's, that would
be beneficial for users reading source.


klauern:
I realized that 3.5 is not installable for black, so I removed it from `.travis.yml`, but applied the formatting to all the files.  If this is too big of a change, I can close it.

j4v:
I like the idea of integrating a formatter to the CI/CD flow. Lack of python 3.5 support is indeed an issue, I don't think we want to drop it at this point.

Would https://github.com/hhatto/autopep8 be an option?

klauern:
I'm for any option, honestly.  Autopep8 is fine by me, too.
-Nick Klauer

j4v:
Tracking this under https://github.com/nccgroup/ScoutSuite/issues/645.

I'll close this now, as Black is considered in "beta" I'd rather not merge in 284 changed files with the current test coverage.

Issue #615, opened 2020-01-23
lack of padding causing exception in providers/aws/facade/ec2.py line 35


rracterr:
```2020-01-23 23:21:38 ip-10-166-140-206.ec2.internal asyncio[11479] ERROR Task exception was never retrieved
future: <Task finished coro=<EC2Instances.fetch_all() done, defined at /home/ec2-user/ScoutSuite/ScoutSuite/providers/aws/resources/ec2/instances.py:14> exception=Error('Incorrect padding')>
Traceback (most recent call last):
  File "/home/ec2-user/ScoutSuite/ScoutSuite/providers/aws/resources/ec2/instances.py", line 17, in fetch_all
    name, resource = await self._parse_instance(raw_instance)
  File "/home/ec2-user/ScoutSuite/ScoutSuite/providers/aws/resources/ec2/instances.py", line 26, in _parse_instance
    instance['user_data'] = await self.facade.ec2.get_instance_user_data(self.region, id)
  File "/home/ec2-user/ScoutSuite/ScoutSuite/providers/aws/facade/ec2.py", line 35, in get_instance_user_data
    value = base64.b64decode(user_data_response['UserData']['Value'])
  File "/usr/lib64/python3.7/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding
```
it appears the code is not handling the missing ==== padding... just a guess.

j4v:
Looks like the user data has invalid padding?

Guess we'll have to handle that case (looks easy enough https://gist.github.com/perrygeo/ee7c65bb1541ff6ac770) and handle exceptions (return `None` if decoding fails).

paurisa:
I added the exception handling and some code to try and decode the User Data by adding extra padding at the end of the Base64 string.
PR #653

paurisa:
We can close this issue now @j4v, right?

PR #653, opened 2020-02-19
Handled Base64 decoding error on EC2 User Data

paurisa:
Handled the Base64 decoding padding error and added some code to try and decode the User Data anyway.
Fixes issue #615

### DIFF
--- a/ScoutSuite/providers/aws/facade/ec2.py
+++ b/ScoutSuite/providers/aws/facade/ec2.py
@@ -32,12 +32,22 @@ class EC2Facade(AWSBaseFacade):
             if 'Value' not in user_data_response['UserData'].keys():
                 return None
             else:
-                value = base64.b64decode(user_data_response['UserData']['Value'] + "===")  # Adds extra padding at the
-                # end to the base64 value to avoid decoding errors
-                if value[0:2] == b'\x1f\x8b':  # GZIP magic number
-                    return zlib.decompress(value, zlib.MAX_WBITS | 32).decode('utf-8')
-                else:
-                    return value.decode('utf-8')
+                try:
+                    value = base64.b64decode(user_data_response['UserData']['Value'])
+                    if value[0:2] == b'\x1f\x8b':  # GZIP magic number
+                        return zlib.decompress(value, zlib.MAX_WBITS | 32).decode('utf-8')
+                    else:
+                        return value.decode('utf-8')
+                except base64.binascii.Error as e:
+                    print_exception(
+                        'Encoded UserData in Base64 has an incorrect padding')
+                finally:
+                    value = base64.b64decode(user_data_response['UserData']['Value'] + "===")  # Adds extra padding at
+                    # the end to the base64 value to avoid decoding errors
+                    if value[0:2] == b'\x1f\x8b':  # GZIP magic number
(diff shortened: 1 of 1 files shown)


ScoutSuite/providers/aws/facade/ec2.py
```diff
@@ -32,11 +32,22 @@ def __init__(self, session: boto3.session.Session, owner_id: str):
...
-                    return zlib.decompress(value, zlib.MAX_WBITS | 32).decode('utf-8')
-                else:
-                    return value.decode('utf-8')
+                try:
+                    value = base64.b64decode(user_data_response['UserData']['Value'])
+                    if value[0:2] == b'\x1f\x8b':  # GZIP magic number
+                        return zlib.decompress(value, zlib.MAX_WBITS | 32).decode('utf-8')
+                    else:
+                        return value.decode('utf-8')
+                except base64.binascii.Error as e:
```
j4v: Do we need to generate an exception, if it's something we can handle silently?

ScoutSuite/providers/aws/facade/ec2.py
```diff
@@ -32,11 +32,22 @@ def __init__(self, session: boto3.session.Session, owner_id: str):
...
+                    print_exception(
+                        'Encoded UserData in Base64 has an incorrect padding')
+                finally:
+                    value = base64.b64decode(user_data_response['UserData']['Value'] + "===")  # Adds extra padding at
+                    # the end to the base64 value to avoid decoding errors
+                    if value[0:2] == b'\x1f\x8b':  # GZIP magic number
```
j4v: Here we're doubling the code. Wouldn't it be better to first validate padding, add it if missing (as described in https://gist.github.com/perrygeo/ee7c65bb1541ff6ac770), and then process the user data once that's done?

paurisa: In my first commit I just added extra padding `===` to the end of every User Data value. This way, the decode is always successful as the extra padding is removed by the decode function if it's not needed.
That would also be my preferred solution however I was not able to add a warning or a disclaimer in case the User Data string was incorrect. I would suggest to add a little disclaimer stating that the decoded data was obtained by adding extra padding at the end of the Base64 encoded string and, therefore, the resulting output could be wrong.

How do you find this solution?

ScoutSuite/providers/aws/facade/ec2.py
```diff
@@ -32,11 +32,19 @@ def __init__(self, session: boto3.session.Session, owner_id: str):
...
-                else:
-                    return value.decode('utf-8')
+                try:
+                    return await self._decode_user_data(user_data_response['UserData']['Value'])
+                except base64.binascii.Error as e:
+                    return await self._decode_user_data(base64.b64decode(user_data_response['UserData']['Value'] + "==="))
+                except Exception as e:
+                    print_exception('Unable to decode EC2 instance user data: {}'.format(e))
+
+    async def _decode_user_data(self, user_data):
```
j4v: @paurisa how does this look? I've moved the decoding to a specific function, and added a retry if we catch an error.

paurisa: That sounds great!

PR #669, opened 2020-03-13
Enhancement/Azure Microsoft Graph ++ Support

j4v:
- [ ] Adds support for a number of APIs
  - [ ] Microsoft Graph
     - Could be useful: https://developer.microsoft.com/en-us/graph/graph-explorer
     - **Update** looks like there's now an official client https://github.com/microsoftgraph/msgraph-sdk-python-core
  - ~~AAD Graph version `1.6-internal`~~
      - Looks like we can't hotpatch the version in the SDK (not sure why)
  - ~~`main.iam.ad.ext.azure.com`~~
- [ ] Implement and test authentication for all modes
  - ~~Some will not work, e.g. `main.iam.ad.ext.azure.com` for Service Principals.~~ 
  - ~~In those cases, we'll have to handle not pulling resources when those authentication modes are chose.~~
  - This will need to be documented.
- [ ] Investigate what additional CIS checks can be supported with these new APIs
- [ ] Update wiki with new configuration requirements for all authentication modes

Edit: we've decided against using non-official authentication modes & endpoints.


j4v:
Leaving this as `blocked` for now. The https://pypi.org/project/azure-identity/ and https://pypi.org/project/msal/ don't seem mature enough and have poor support for username/password flows, and the Microsoft Graph API doesn't seem to provide any additional information.

j4v:
Closing this PR as the work is being done as part of https://github.com/nccgroup/ScoutSuite/pull/1026 and associated branches.

Issue #467, opened 2019-07-16
Feature Request - List supported services


museadmin:
Hi

You support skipping services in AWS via these options:

  --services SERVICES [SERVICES ...]
                        Name of in-scope services.
  --skip SKIPPED_SERVICES [SKIPPED_SERVICES ...]
                        Name of out-of-scope services.

Can you add an option to list supported services that you can skip or include?

My thinking is that it would make it simple for the user to use the correct service name format when using the above options. Also, as people add in services, it could dynamically update the list of services available to skip or include.

j4v:
Thanks @museadmin, I agree it would be a good improvement. A bit tricky to implement as services are loaded dynamically but I'm sure we can figure it out.

j4v:
> A bit tricky to implement as services are loaded dynamically but I'm sure we can figure it out. 

Not that tricky if we check this dynamically though...

j4v:
Closing, implemented in `develop`.

PR #671, opened 2020-03-16
Feature/Added --list-services argument.

JJmako:
Added `--list_services` argument, to resolve https://github.com/nccgroup/ScoutSuite/issues/467.

With this option, all possible services are listed.


j4v:
@JJmako the current implementation requires authentication. would it be possible to implement it similarly to `--version`?

JJmako:
> @JJmako the current implementation requires authentication. would it be possible to implement it similarly to `--version`?

I have been chasing this. You need to authenticate to retrieve the services. If I do a parameter similar to --version I have to mock the services in a file.
If the services are always the same, this is a good solution, but if providers change their services this file have to change.

j4v:
> You need to authenticate to retrieve the services. 

I believe you're right. I thought you could just instantiate a facade or ServiceConfig object and look at its attributes, but those require credentials to build. 

Therefore I think your approach is correct, I'll review and merge :+1:

Issue #434, opened 2019-06-03
Update AWS CIS Benchmark Ruleset


j4v:
- [ ] Review the existing ruleset & findings to ensure everything in v1.2.0 is implemented
  - Open GH issues for missing findings
- [ ] Update ruleset name to `cis-1.2.0.json`

j4v:
Closing, this is done in `develop`.

PR #679, opened 2020-03-19
Enhancement/Update AWS CIS Benchmark to 1.2.0

paurisa:
Changes:
- Creates the new AWS CIS 1.2.0 json ruleset, the required new findings and updates the findings to the new output format
  - Including 23 new rules
- Adds support for AWS CloudWatch Metric Filters
- Adds support for AWS Peering Connections
- Adds support for AWS Flow Logs (Subnet & VPC)

This PR is for issue #434 

Phases:

- [x]  Create all the new AWS CIS 1.2.0 IAM rules and update the format of the old ones
- [x]  Create all the new AWS CIS 1.2.0 Logging rules and update the format of the old ones
- [x]  Create all the new AWS CIS 1.2.0 Monitoring rules and update the format of the old ones
- [x]  Create all the new AWS CIS 1.2.0 Networking rules and update the format of the old ones

TODOs for @j4v:
- [x] Complete finding 1.14 of the `cis-1-2-0.json` ruleset
This finding is already created in the rule `iam-root-account-no-hardware-mfa.json`. The resource `iam/credentialreports.py` is already set to fetch the `VirtualMfaDevices` and parse it. The parsing of the returned data should get the serial-number of the mfa device and also a boolean containing whether the MFA device is hardware based or not.
To check if a MFA device is hardware or virtual, the serial-number should be checked. If the serial-number contains "arn", then the device is virtaul. Otherwise, the device is hardware.
Once done, issue #681 can be closed.
- [x] Ensure there aren't any duplicate findings in pub/prop
  - [x] Compare EC2 SG findings as there might be duplicates
- [x] Ensure new findings are added to `default` ruleset (+prop) where appropriate
- [x] Ensure new findings are added to `detailed` ruleset where appropriate
  - [x] Review the arguments used in the `detailed` ruleset
     There are some findings that may have inappropriate arguments. For example, the `iam-password-policy-minimum-length` finding has the minimum length argument set to `8` when the CIS recommendation is `14`.
- [x] Review peering connection implementation/partial
- [x] Review hidden flow logs implementation/partial
- [x] Fix findings 3.1 to 3.14 of `cis-1-2-0.json` ruleset not returning an empty list when clicked.
  Further information of this issue can be found here: https://github.com/nccgroup/ScoutSuite/pull/679#issuecomment-633639978


ScoutSuite/providers/aws/rules/rulesets/cis-1.2.0.json
```diff
@@ -0,0 +1,164 @@
+{
+    "about": "This ruleset attempts to cover as many recommendations from the CIS Amazon Web Services Foundation v1.0.0.",
```
j4v: `v1.2.0`

ScoutSuite/providers/aws/rules/findings/iam-root-account-used-recently.json
```diff
@@ -1,7 +1,15 @@
 {
     "description": "Root account used recently",
-    "path": "iam.credential_reports.id",
+    "rationale": "<b>Description:</b><br><br>The use of the root account should be avoided.",
```
j4v: Remove `<b>Description:</b><br><br>`

ScoutSuite/providers/aws/rules/findings/iam-user-without-mfa.json
```diff
@@ -1,7 +1,15 @@
 {
     "description": "User without MFA",
-    "path": "iam.users.id",
+    "rationale": "<b>Description:</b><br><br>All IAM users should have Multi Factor Authentication (MFA) enabled.",
```
j4v: Remove `<b>Description:</b><br><br>`

ScoutSuite/providers/aws/rules/findings/iam-root-account-used-recently.json
paurisa: This is fixed and will be available in the next commit.

ScoutSuite/providers/aws/rules/findings/iam-user-without-mfa.json
paurisa: This is fixed and will be available in the next commit.

ScoutSuite/providers/aws/rules/rulesets/cis-1.2.0.json
paurisa: Version updated, will be available in the next commit.

paurisa:
Updated all the old CIS Benchmark 1.0.0 rules to the new format and added all the new CIS Bechmark 1.2.0 rules.

There are still a couple of issues to fix or check but it should all work with some minor tweaks.

paurisa:
I just pushed a new commit where I updated all the old CIS Benchmark 1.0.0 Logging rules to the new format and added all the new CIS Bechmark 1.2.0 Logging rules.

There are a few minor issues to be fixed before it works as intended.

paurisa:
Current errors to be fixed:

- [ ]  iam_unusued_credentials_not_disabled. Check the error on condition 'olderThan' for values where last_used is NoneType.
- [ ]  iam-no-support-role. Is this the correct way to check if the policy AWSSupportAccess (created by default) is attached to at least 1 user? 
- [ ]  iam-managed-policy-allows-full-privileges. Fix error: list index out of range.
- [ ]  cloudtrail-no-logging. Need to add an extra verification (recommended by CIS): Condition to verify there is at least one Event Selector for a Trail with IncludeManagementEvents set to true and ReadWriteType set to All
- [ ]  Recommendation 2.3. Should we create a new finding exclusive to CloudTrail S3 buckets or reuse the already created ones?
- [ ]  Recommendation 2.6. Should be specific to CloudTrail S3 buckets, currently the cis 1.2.0 ruleset has a general bucket logging finding in this recommendation. Finish finding cloudtrail-s3-bucket-no-logging to verify the bucket has the principal cloudtrail.amazonaws.com and logging disabled.
- [ ]  vpc-subnet-without-flow-log. Would like to make the flow log counter higllighted in red when there are 0 flow logs (services.vpc.regions.id.vpcs.id.subnets.html partial)
- [ ]  Recommendation 4.4 ec2-routes-tables-full-peering. Complete finding conditions with the newly created object of VPCs route_tables. Also modify description to better explain the end user that this finding depends on their requirements for each instance

j4v:
@paurisa merged `develop` branch into this one and resolved conflicts.

I fixed the syntax for these, please review they actually flag what they should:
- iam-managed-policy-allows-full-privileges
- iam_unusued_credentials_not_disabled

Also fixed:
- iam-no-support-role
  - Was as bit tricky as we only fetch policies which are attached to principals, so I had to check for it's absence

j4v:
@paurisa looks like there will be some conflicts due to https://github.com/nccgroup/ScoutSuite/pull/711 but they should be easy enough to resolve.

paurisa:
Will look into it. Thanks for the heads up!

paurisa:
Implemented template for CIS 1.2.0 benchmark Monitoring rules and resolved merge conflicts.

j4v:
@paurisa I've added the metric filters required for CloudWatch rules.

I've also written the https://github.com/nccgroup/ScoutSuite/blob/95f9710faa25ebb79320219419725869fc9c8b50/ScoutSuite/providers/aws/rules/findings/logs-no-alarm-signin-without-mfa.json rule. **Note** that this was a bit tricky, as we have to check for each region if there is "one" metric filter that checks the desired event. To implement this I've had to add the validation in the `finalize` method of the resource, see  https://github.com/nccgroup/ScoutSuite/blob/95f9710faa25ebb79320219419725869fc9c8b50/ScoutSuite/providers/aws/resources/cloudwatch/base.py#L17.

@alessandrogonzalez do you think there's a better way to do this? Looks a lot like the issues we had with CloudTrail.

alessandrogonzalez:
> @paurisa I've added the metric filters required for CloudWatch rules.
> 
> I've also written the https://github.com/nccgroup/ScoutSuite/blob/95f9710faa25ebb79320219419725869fc9c8b50/ScoutSuite/providers/aws/rules/findings/logs-no-alarm-signin-without-mfa.json rule. **Note** that this was a bit tricky, as we have to check for each region if there is "one" metric filter that checks the desired event. To implement this I've had to add the validation in the `finalize` method of the resource, see
> 
> https://github.com/nccgroup/ScoutSuite/blob/95f9710faa25ebb79320219419725869fc9c8b50/ScoutSuite/providers/aws/resources/cloudwatch/base.py#L17
> 
> .
> @alessandrogonzalez do you think there's a better way to do this? Looks a lot like the issues we had with CloudTrail.

I have not tested the code but I think that this will highlight all regions that do not have a cloudwatch rule for login without MFA. Should this only be raised when no region has a rule like this? (something like this: https://github.com/nccgroup/ScoutSuite/blob/develop/ScoutSuite/providers/aws/resources/cloudtrail/base.py#L32)

j4v:
> Should this only be raised when no region has a rule like this?

Good question, I implemented it this way as I thought the CloudWatch events where per-region, but I guess if the CloudTrail that triggers the even is per-account then in this case we won't want to check it for each region.

One issue is that when you click on the rule it returns an empty (blank) list of affected objects, I'm not sure why.

alessandrogonzalez:
> One issue is that when you click on the rule it returns an empty (blank) list of affected objects, I'm not sure why.

I think it is because the `path` attribute of the rule should be `cloudwatch.regions.id.metric_filters.id` and the `display_path` can be removed. Otherwise it will try to use the template of a `metric_filter`, but it will not have the `metric_filter` details to render the template.

paurisa:
I implemented all the alarm findings in the `ScoutSuite/ScoutSuite/providers/aws/resources/cloudwatch/base.py`
I also changed the rules to the logical operator "or" instead of "and". My reasoning behind it is that if the alarm is active in a region, then it is therefore available in all regions as the trail associated with it should be MultiRegion.

There is also the fact that the findings return an empty list when clicked. I tried what @alessandrogonzalez suggested but it always returns the same text for all the findings. Will have to look more into it.

Finally, we should create a battery of tests to check that all alarm findings are triggered correctly. As of now, I have only been able to test a single alarm although they should all work correctly.

j4v:
Finally ready to merge! Thanks @paurisa for the great work on this.

paurisa:
No problem, it is a pleasure to contribute to this useful project.

PR #733, opened 2020-05-05
Feature/AWS Standard ARN Keys

### LINKED ISSUE #511: Proposal for having ARN as a standard key across all service assets being scanned for AWS
Currently different service in scoutsuite have different set of key (non-standard) attributes depending on the nature of service and the data required to identify a finding.

This issue proposes to have one more standard key  called 'arn'  containing the corresponding asset's arn value over and above already existing attributes for each and every service asset.  Some services do have arn as part of their attribute set, but not all. 

Example : 'iam' has arn for all of it's assets except credential report (which obviously would not make sense). But 'ec2' assets such as snapshots , security_groups etc does not have arn as part of their attribute. 

Though currently 'id' exists as a unique value across all service assets, this is not human readable as arn.  Arn would give out the region, service , account_id , aws_partition and the asset (all in one key). It will be easier to relate in case any other tool / script which depends on scoutsuite results would want to give out a report with list of assets and it's corresponding issues.

JJmako:
# Description

Add ARN processing and show in HTML report for:

- [x] IAM
  - [x] groups
  - [x] users
- [x] Awslambda
  - [x] function
- [x] Cloudformation
  - [x] stack
- [x] EC2
  - [x] snapshot
  - [x] volume
  - [x] image
  - [x] instance
  - [x] network_interface
  - [x] security_group
- [x] elasticache
  - [x] cluster
- [x] ELB
  - [x] loadbalancer
- [x] RDS
  - [x] instance
- [x] Redshift
  - [x] cluster
  - [x] parameter_group
- [x] Route53 
  - [x] domain
- [x] S3
  - [x] bucket
- [x] SES
  - [x] identity
- [x] VPC
  - [x] vpc

Following the nomenclature from:
https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

Fixes #511 

## Type of change

## Checklist:


j4v:
@JJmako updated description to make it easier to track progress. Also don't forget to add the fields to the HTML partials.

ScoutSuite/output/data/html/partials/aws/services.ec2.regions.id.snapshots.html
```diff
@@ -7,7 +7,8 @@ <h4 class="list-group-item-heading">{{name}}</h4>
     <div class="list-group-item">
         <h4 class="list-group-item-heading">Information</h4>
         <ul>
-            <li class="list-group-item-text">Id: {{id}}</li>
+            <li class="list-group-item-text">ID: {{id}}</li>
+            <li class="list-group-item-text">ARN: {{arn}}</li>
```
j4v, line 11: @JJmako here's an example of adding the ARN to the "partial" that gets displayed in the report. All you need to do once the ARN is implemented in the parsing is add a line like this one to the partial so it shows up in the report.

There are a few partials (like the one for elasticache clusters - https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/output/data/html/partials/aws/services.elasticache.regions.id.vpcs.id.clusters.html) which uses an old generic template, for those you won't need to do anything.

j4v:
@lowSoA please merge `develop` into this branch to resolve conflicts.

ScoutSuite/providers/aws/resources/ec2/snapshots.py
```diff
@@ -18,6 +18,9 @@ def _parse_snapshot(self, raw_snapshot):
         raw_snapshot['id'] = raw_snapshot.pop('SnapshotId')
         raw_snapshot['name'] = get_name(raw_snapshot, raw_snapshot, 'id')
         raw_snapshot['public'] = self._is_public(raw_snapshot)
+        raw_snapshot['arn'] = 'arn:aws:ec2:{}:{}:snapshot/{}'.format(self.get('region'),
+                                                                             self.facade.owner_id,
```
alessandrogonzalez, line 22: I think that, as snapshots could be public, in this case it would be better to get the owner ID from the raw_snapshot object

j4v: Agreed - I'd think this is included in the `raw_snapshot`?

Issue #438, opened 2019-06-06
Export finding summary


j4v:
Currently you can export details of resources affecting a finding. It would be nice to also be able to export a summary of findings (e.g. to share within an org).

Do this for:
- [ ] CSV
- ~~PDF~~
- ~~HTML~~

j4v:
Closing, this has been implemented in `develop`.

PR #752, opened 2020-05-22
Feature/438 export finding summary

JJmako:
Issue related to https://github.com/nccgroup/ScoutSuite/issues/438

It exports a CSV with the summary of findings.


j4v:
@JJmako tested and it works perfectly. The only thing I'd add is a column with the risk level (danger/warning) - is that an easy fix?

JJmako:
> @JJmako tested and it works perfectly. The only thing I'd add is a column with the risk level (danger/warning) - is that an easy fix?

Added! Was an easy fix :)

PR #766, opened 2020-06-03
Include more s3 information

alessandrogonzalez:
# Description

This PR adds information about public access block configuration to S3 buckets.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works (optional)

### DIFF
--- a/ScoutSuite/providers/aws/facade/s3.py
+++ b/ScoutSuite/providers/aws/facade/s3.py
@@ -219,7 +219,7 @@ class S3Facade(AWSBaseFacade):
     def get_s3_public_access_block(self, account_id):
         # We need a region to generate the client
         # However, the settings are global, so they are not region-dependent
-        region = 'eu-west-1'
+        region = 'us-east-1'
         client = AWSFacadeUtils.get_client('s3control', self.session, region)
         try:
             s3_public_access_block = client.get_public_access_block(AccountId=account_id)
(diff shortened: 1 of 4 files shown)


ScoutSuite/providers/aws/facade/s3.py
```diff
@@ -204,6 +216,27 @@ def _set_s3_bucket_secure_transport(self, bucket: {}):
             print_exception('Failed to evaluate bucket policy for %s: %s' % (bucket['Name'], e))
             bucket['secure_transport'] = None
 
+    def get_s3_public_access_block(self, account_id):
+        # We need a region to generate the client
+        # However, the settings are global, so they are not region-dependent
+        region = 'eu-west-1'
```
ramimac, line 222: I haven't tested this out yet, so just dropping a quick not.
Hard-coding regions has bit us repeatedly in the past, so definitely an anti-pattern. 
At the very least, we should use `us-east-1`, as it can't be disabled - see https://github.com/nccgroup/ScoutSuite/pull/705

j4v: See https://github.com/nccgroup/ScoutSuite/blob/develop/ScoutSuite/providers/aws/facade/s3.py#L14, which will take the region configured in the session if non is provided.

PR #791, opened 2020-06-12
Simplify Rule Mapping - include all fields in rule.json

klauern:
# Description

All of the fields in a rule are not contained in the ScoutSuite Results `.js` file.  This change simplifies the mapping by including all of the fields in a rule, including arguments and conditions.  These fields can then be used downstream for consumption either as part of a translation layer, or included into the UI for better context mapping.

## Type of change

## Checklist:


j4v:
I'm not sure this is something we want to include by default. Could do so when running with `--debug` flag though.

klauern:
That's fair.  Part of the reasoning behind this is that we want to be flexible about what information is going to end up in the raw results.  As it is in JSON format, it was easy enough for us to just pick out the fields we want and ignore the rest.  In it's current form, it only considers fields explicitly, and requires a code change for every new field or mapping that ends up.

On a separate note, we also want to be able to pull in arbitrary fields from the rulesets, as those are more customizable than the source-based rule definitions, and it does not appear that there's a mapping from the ruleset to the result format, either.  We would like to be able to add our own additional fields to the rulesets, but it doesn't appear that any of those fields make it to the resulting format.

ScoutSuite/core/processingengine.py
```diff
@@ -40,38 +37,30 @@ def run(self, cloud_provider, skip_dashboard=False):
...
+                rule_dict = {"description": rule.description, "path": rule.path}
                 for attr in ['level', 'id_suffix', 'class_suffix', 'display_path']:
                     if hasattr(rule, attr):
-                        cloud_provider.services[service][self.ruleset.rule_type][rule.key][attr] = getattr(rule, attr)
+                        rule_dict[attr] = getattr(rule, attr)
                 try:
                     setattr(rule, 'checked_items', 0)
-                    cloud_provider.services[service][self.ruleset.rule_type][rule.key]['items'] = recurse(
-                        cloud_provider.services, cloud_provider.services, path, [], rule, True)
+                    for a in dir(rule):
```
alessandrogonzalez, line 47: Wouldn't this loop include the attributes set in the previous for loop? (lines 42-44)

ScoutSuite/core/processingengine.py
```diff
@@ -40,38 +37,30 @@ def run(self, cloud_provider, skip_dashboard=False):
...
+                    rule_dict['flagged_items'] = len(cloud_provider.services[service][self.ruleset.rule_type][rule.key]['items'])
                 except Exception as e:
                     print_exception('Failed to process rule defined in %s: %s' % (rule.filename, e))
                     # Fallback if process rule failed to ensure report creation and data dump still happen
-                    cloud_provider.services[service][self.ruleset.rule_type][rule.key]['checked_items'] = 0
-                    cloud_provider.services[service][self.ruleset.rule_type][rule.key]['flagged_items'] = 0
+                    rule_dict['checked_items'] = 0
+                    rule_dict['flagged_items'] = 0
+
+                cloud_provider.services[service][self.ruleset.rule_type][rule.key] = rule_dict
```
alessandrogonzalez, line 63: I **think** that with this assignment you would be potentially losing the info stored at cloud_provider.services[service][self.ruleset.rule_type][rule.key]['items'], unless this is copied to rule_dict somewhere I have not seen

ScoutSuite/core/processingengine.py
klauern: man, now you're testing me.  lol.  

I believe I removed the lines in 43-45 (https://github.com/nccgroup/ScoutSuite/pull/791/files#diff-3bd3294a5da787ffbbd439e6ac8f6a1aL43-L45) as the looping in this group was inclusive of those, so it didn't need to be perforemd.

ScoutSuite/core/processingengine.py
klauern: Ah, yeah, you're probably right.  I wanted this to be an `UPSERT` (for lack of a better term) of the existing dictionary.  I know that I am running this and getting some inconsistent results so this might be why...

j4v:
@klauern is this still relevant? If not please close the PR.

j4v:
Closing, happy to reopen/discuss if further work is done towards this.

Issue #838, opened 2020-08-29
Feature Request : Include tag information for IAM Users


rams3sh:
**Is your feature request related to a problem? Please describe.**
User Resource lacks tags.

**A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]**

There are some use case for me where certain tags are exempted from a compliance rule and others are not.
Due to lack of tags information being captured as part of IAM user resource, am unable to implement such rule checks.

**Describe the solution you'd like**
Expected output for iam users in result json :- 

**Describe alternatives you've considered**

I use a custom stand alone script to audit such cases alone separately.

**Additional context**

None

rams3sh:
Missed the expected output in the original issue.

Below the expected json output :- 
![image](https://user-images.githubusercontent.com/5143597/91644746-a4bd9480-ea5c-11ea-91e5-11a99c4c210f.png)

j4v:
Closing, this is implemented in `develop`.

PR #843, opened 2020-09-02
Feature/include tag

JJmako:
PR related to https://github.com/nccgroup/ScoutSuite/issues/838

Added tags to IAM users.

### DIFF
--- a/ScoutSuite/providers/aws/facade/iam.py
+++ b/ScoutSuite/providers/aws/facade/iam.py
@@ -99,6 +99,7 @@ class IAMFacade(AWSBaseFacade):
         await get_and_set_concurrently(
             [functools.partial(self._get_and_set_inline_policies, iam_resource_type='user'),
              self._get_and_set_user_groups,
+             self._get_and_set_user_tags,
              self._get_and_set_user_login_profile,
              self._get_and_set_user_access_keys,
              self._get_and_set_user_mfa_devices],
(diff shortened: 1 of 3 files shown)


ScoutSuite/providers/aws/facade/iam.py
```diff
@@ -99,6 +99,7 @@ async def get_users(self):
         await get_and_set_concurrently(
             [functools.partial(self._get_and_set_inline_policies, iam_resource_type='user'),
              self._get_and_set_user_groups,
+             self._get_and_set_user_tags,
```
ramimac, line 102: Is this necessary? My understanding is that [list_users already contains the Tags](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.Client.list_users), we shouldn't have to make all these additional calls?

ScoutSuite/providers/aws/resources/iam/users.py
```diff
@@ -16,5 +16,6 @@ def _parse_user(self, raw_user):
         raw_user['id'] = raw_user.pop('UserId')
         raw_user['name'] = raw_user.pop('UserName')
         raw_user['arn'] = raw_user.pop('Arn')
-
+        if (len(raw_user['tags']['Tags']) > 0):
```
ramimac, line 19: If you don't need _get_and_set_user_tags, tags will be accessible directly in raw_user['Tags']

JJmako: I was triying to access ['Tags'] from HTML, but without this conversion was imposible for me

ScoutSuite/providers/aws/facade/iam.py
JJmako: I had to call to "list_user_tags" for getting the tags. Without that call, the user don't contain the Tags.

lowSoA: I've run a quick test and can't see any tags being returned. I've looked into list_users with boto3 and also with the CLI (list-users) and same results, no tags, even though it's in the documentation as you've mentioned (https://docs.aws.amazon.com/cli/latest/reference/iam/list-users.html).

ramimac (approved):

j4v:
I've modified the partial so it looks like the other Tags / resources.

Issue #570, opened 2019-11-13
enhancement/Azure SG rules


j4v:
Currently we only check if SSH/RDP/SQL ports are open to the Internet. Add 2-3 rules similar to the ones we have for AWS.

j4v:
This has been implemented in `develop`.

PR #848, opened 2020-09-08
Improve implementation of Azure NSG port ranges

### LINKED ISSUE #793: Bug: Poor implementation of Azure NSG ports
See https://github.com/nccgroup/ScoutSuite/issues/698#issuecomment-645229279.

Need to improve this implementation.

lowSoA:
# Description

- Addressed the issue where port ranges were parsed into a long list of ports, specially the default inbound/outbound rules
- As a result, the final JS file is much smaller which could affect host performance if many default rules are in place

Fixes #793

## Type of change

## Checklist:

### DIFF
--- a/ScoutSuite/core/conditions.py
+++ b/ScoutSuite/core/conditions.py
@@ -212,6 +212,7 @@ def pass_condition(b, test, a):
 
     # Port/port ranges tests
     elif test == 'portsInPortList':
+        result = False
         if not type(b) == list:
             b = [b]
         if not type(a) == list:
@@ -224,12 +225,13 @@ def pass_condition(b, test, a):
                     if type(port) != int:
                         port = int(port)
                     if bottom_limit_port <= port <= upper_limit_port:
-                        return True
+                        result = True
+                        break
             else: #A single port
                 for port in a:
                     if port == port_range:
-                        return True
-        return False
+                        result = True
+                        break
 
     # Policy statement tests
     elif test == 'containAction':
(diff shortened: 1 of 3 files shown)


j4v:
Great! Does https://github.com/nccgroup/ScoutSuite/blob/3fde85ee3c36067ec8f35e84eecc83d2cb8df19e/ScoutSuite/providers/azure/rules/findings/network-security-groups-rule-inbound-internet-all.json also require an update?

j4v:
If you have the time, it would be good to check if we can add any rules as described in https://github.com/nccgroup/ScoutSuite/issues/570, since it relates to the same functionality.

lowSoA:
> Great! Does https://github.com/nccgroup/ScoutSuite/blob/3fde85ee3c36067ec8f35e84eecc83d2cb8df19e/ScoutSuite/providers/azure/rules/findings/network-security-groups-rule-inbound-internet-all.json also require an update?

It does not need an update since looks for `*` or `0-65535` in `destination_port_ranges` parameter and the new update was primarily around the `destination_ports` parameter, which affects the `network-security-groups-rule-inbound-service.json` rule directly.

I've had `destination_port_ranges` parameter with the rule you mentioned in mind but it doesn't seem to need an update (IMO).

> If you have the time, it would be good to check if we can add any rules as described in #570, since it relates to the same functionality.

I'll look into this.

j4v:
Brilliant! It makes a huge difference, even for a small subscription:

```sh
(venv) ~/Git/ScoutSuite (develop ✘)✭ ᐅ l [report]

-rw-rw-r-- 1 user user 5.9M Sep 19 10:29 [report]

(venv) ~/Git/ScoutSuite (bug/793-azure-nsg-reports ✘)✭ ᐅ l [report]

-rw-rw-r-- 1 user user 20K Sep 19 10:29 [report]
```

PR #849, opened 2020-09-09
Improved AWS default SG rules

### LINKED ISSUE #5: Improvable Default SG Rules
`ec2-default-security-group-in-use.json`
- Checks that the `default` SG is in use, but not that the SG hasn't been modified to be secure.
- Add a `is_default_configuration` parameter to validate.

`ec2-default-security-group-with-rules.json`:
- Check that the `default` SG is in use and that is isn't empty, but not that it has the default configuration. Apply same check as previous.

lowSoA:
# Description

- Added `is_default_configuration` SG parameter based on the following ([source](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-security-groups.html)):

```
A default security group is named default, and it has an ID assigned by AWS. The following are the default rules for each default security group:

- Allows all inbound traffic from other instances associated with the default security group. The security group specifies itself as a source security group in its inbound rules.

- Allows all outbound traffic from the instance.
```

- Added previous parameter check to the following rules:
  - `ec2-default-security-group-in-use.json`
  - `ec2-default-security-group-with-rules.json`

Fixes #5

## Type of change

## Checklist:

### DIFF
--- a/ScoutSuite/providers/aws/resources/ec2/securitygroups.py
+++ b/ScoutSuite/providers/aws/resources/ec2/securitygroups.py
@@ -41,7 +41,29 @@ class SecurityGroups(AWSResources):
             raw_security_group['IpPermissionsEgress'])
         security_group['rules']['egress']['protocols'] = egress_protocols
         security_group['rules']['egress']['count'] = egress_rules_count
+
+        if self._has_default_egress_rule(raw_security_group['IpPermissionsEgress']) and self._has_default_ingress_rule(raw_security_group['IpPermissions'], raw_security_group['GroupId']):
+            security_group['is_default_configuration'] = True
+        else:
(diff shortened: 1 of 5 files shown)


lowSoA:
I'm not familiar with tests on GitHub, but my guess is that the tests are not accurate since they only check what is mentioned in #5: name, if being used and if not empty. 

But ultimately, I _think_ the checks failed because there is no `is_default_configuration` parameter in the test cases/rule-configs since this is a new addition.

j4v:
> Added is_default_configuration SG parameter based on the following (source):

Very good implementation :medal_military: 

> But ultimately, I think the checks failed because there is no is_default_configuration parameter in the test cases/rule-configs since this is a new addition.

Yep, fixed it.

PR #852, opened 2020-09-11
DELETE ME :(

shanedroid:
Delete Me :(


j4v:
?

ramimac:
@j4v they have a fork they meant this PR for as far as I can tell. Unfortunately there is no way to delete PRs, but I don't see anything "bad" in these artifacts, so no harm really in leaving it closed.

shanedroid:
> @j4v they have a fork they meant this PR for as far as I can tell. Unfortunately there is no way to delete PRs, but I don't see anything "bad" in these artifacts, so no harm really in leaving it closed.

Yep. If you would be so kind as to delete this PR, it would be much appreciated. 🥺

shanedroid:
at least I learned something 📚  

https://sethrobertson.github.io/GitFixUm/fixup.html

Issue #873, opened 2020-09-23
Improve error handling across all resources


j4v:
Currently, facades implement good exception handling, but resource parsing does not. That means that for a given resource type, if parsing fails for any given resource, the `fetch_all` method fails and stops, hence not parsing any additional resources.

All resources should be reviewed and updated, to ensure they handle parsing errors.

e.g., for AWS IAM roles (https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/aws/resources/iam/roles.py#L8) the method should be updated as so:

```python
for raw_role in raw_roles:
    try:
        name, resource = self._parse_role(raw_role)
        self[name] = resource
    except Exception as e:
        print_exception('Failed to parse {} resource: {}'.format(self.__class__.__name__, e))
```

Aamir017:
Hi, I’m looking into improving error handling during resource parsing so that failures in individual resources don’t stop fetch_all() for the entire provider.

Before working on a PR, I wanted to ask:

1.  Is this issue still relevant with the current codebase?

2.  Would you prefer a centralized/shared parsing helper (to handle exceptions consistently across resources), or incremental fixes per resource starting with high-impact ones (e.g., AWS IAM / EC2)?
 
3. Are there any logging or error-reporting conventions you’d like contributors to follow?

Happy to align with the project’s preferred approach before implementing anything.

aadelb:
Hi, I'd like to work on this issue.

**My approach:**
- Audit current error handling across AWS, Azure, and GCP providers
- Implement consistent exception handling patterns
- Add proper logging for debugging
- Ensure graceful degradation when API calls fail

**Timeline:**
I can start with one provider (AWS) and have a PR ready within 1-2 weeks, then extend to others.

Let me know if you'd like me to proceed!

---
Ahmed Adel Bakr Alderai

PR #908, opened 2020-10-21
Added Error handling to every resource

JJmako:
Issue related: https://github.com/nccgroup/ScoutSuite/issues/873

Added Try/Except to every fetch_all functions.


j4v:
@JJmako you can use a `set()` to store the unique errors generated by parsing in the `for` loop. At the end of the loop you can then output the unique errors. 

What we could do if there are say more than 10 unique parsing errors for a single resource is to only display 10 and also output something like "<x> additional errors not displayed", otherwise we may end up outputting hundreds of very similar errors.

j4v (changes requested):
@JJmako you can use a `set()` to store the unique errors generated by parsing in the `for` loop. At the end of the loop you can then output the unique errors. 

What we could do if there are say more than 10 unique parsing errors for a single resource is to only display 10 and also output something like "<x> additional errors not displayed", otherwise we may end up outputting hundreds of very similar errors.

PR #945, opened 2020-12-06
GCP: Always check IAM policies

### LINKED ISSUE #943: GCP scanning ignores IAM service if IAM API not enabled in the target project
@j4v filling out yet another bug to cover this IAM thing I mentioned in #940 - I haven't quite got this one straight in my head, which I usually look to do _before_ filing bugs, so please bear with me while I spam the issue tracker!  Thank you for all the work you do on this. :)

**Describe the bug**

When scanning GCP projects which do not have the IAM API enabled, no IAM results are reported.  However, we know all projects have IAM roles.

When I comment out the Service Usage API check, I get results reported for all projects, regardless of whether the IAM API is enabled in that project.  This may be because I'm running ScoutSuite in its own GCP project (in a separate organization) so it always has access to the IAM API?

This shows that it is possible to scan IAM permissions without the IAM API being enabled for the target project.

In general, I am concerned that just because a service API is not enabled for a project, this does not mean that that service is not used by that project - potentially this might extend beyond IAM.

**To Reproduce**

Scan a project without the IAM API enabled.  Observe no results.

Comment out the service usage API checks in facade/base.py.  Run the scan again and get results.  (May depend on running ScoutSuite from another project that has the IAM API enabled?)

timretout:
# Description

The status of the IAM API has no bearing on whether a project uses IAM.

Note this change might cause errors if the project where ScoutSuite runs does not have the IAM API enabled.

Fixes #943.

## Type of change

## Checklist:

### DIFF
--- a/ScoutSuite/providers/gcp/facade/base.py
+++ b/ScoutSuite/providers/gcp/facade/base.py
@@ -130,6 +130,10 @@ class GCPFacade(GCPBaseFacade):
         Given a project ID and service name, this method tries to determine if the service's API is enabled
         """
 
+        # All projects have IAM policies regardless of whether the IAM API is enabled.
+        if service == 'IAM':
+            return True
+
         serviceusage_client = self._build_arbitrary_client('serviceusage', 'v1', force_new=True)
         services = serviceusage_client.services()
         try:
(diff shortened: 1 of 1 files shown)


j4v:
> Note this change might cause errors if the project where ScoutSuite runs does not have the IAM API enabled.

Did you confirm this? Ideally we'd want to suppress these error messages, since they are FPs and may cause some confusion.

PR #1019, opened 2021-02-08
Feature/979 overview dashboard

SamuelMeilleur:
# Description

Draft PR for the Home Dashboard
Fixes # 979

## Type of change

## Checklist:

- [x] New and existing unit tests pass locally with my changes


ScoutSuite/frontend/src/components/TabsMenu/index.js
```diff
@@ -0,0 +1,38 @@
...
+    tabs, 
+    selectedTab,
+    onClick,
+  } = props;
+
+  return (
+    <div className="tabs-menu">
+      {Object.entries(tabs).map(([key, tabName]) => (
+        <span 
+          className={cx('tab', {'is-selected': selectedTab === tabName})}
```
SamuelMeilleur, line 24: Ideally, I would've preferred to use the object keys rather than names to check which tab is selected but I didn't like the idea of having `Object.keys` everywhere. Maybe there's another way to do it cleanly that I didn't think of.

loic294: I personally think that at component based solution would be better and easier to understand. Here is an example of how ant.design does it: https://ant.design/components/tabs/

Basically, the component doesn't need to take any props other than the childrens (a list of tabs). Each child component can have the required name and potentially an key prop (tho we could generate it automatically based on the index and always display the first one by default). Here is how to read the content props of the childrens: https://stackoverflow.com/questions/29670173/accessing-react-component-children-props-from-the-parent

This also avoids having to create the switch, the tabsName object, etc.

Let me know what you think of this solution!

SamuelMeilleur: I updated the component following your recommendations. It's much simpler to use now.

PR #1223, opened 2021-03-30
GCP Network -DNS zones CISv3.3 to 3.5

xnkevinnguyen:
# Description
New Facade and ressource DNS since this is a separate Google Discovery Endpoint. I've implemented the rules to check if DNSSEC is enabled, and if it is enabled that the encryption algorithms is not RSASHA1 (Weak algorithm).

## Type of change

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code


j4v (changes requested):
For new GCP services, you'll want to first confirm the API for that service is enabled here https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/gcp/facade/base.py#L135, otherwise you'll get a bunch of errors when it isn't.

Looks good otherwise.

xnkevinnguyen:
> For new GCP services, you'll want to first confirm the API for that service is enabled here https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/gcp/facade/base.py#L135, otherwise you'll get a bunch of errors when it isn't.
> 
> Looks good otherwise.

@j4v I've just added the verification to check if the api is enabled.

j4v (approved):

PR #1313, opened 2021-06-03
Fix secret rule which create noise

shivamkumar001:
Fixes # (1028)
-  remove 'secret' from aws ec2 instance.py


x4v13r64:
This does nothing. We still want to catch all potential secrets, just not include them in the report/data on disk.

PR #1269, opened 2021-04-13
Enhancement/Improve ARN processing performance

lowSoA:
# Description

After implementing PR #1022, a large number of calls were made to [get_partition_name()](https://github.com/nccgroup/ScoutSuite/blob/3be326d0b6b14ea67f9cfcd118a726eb5f31e811/ScoutSuite/providers/aws/utils.py#L19). The following table shows a comparison between the original number of calls, the merged changes and the proposed changes (output merged from different Python's cProfiles):

| ncalls | tottime | percall | cumtime | percall | filename:lineno(function) | Branch |
|:------:|:-------:|:-------:|:-------:|:-------:|:-------------------------:|:------:|
| 25 | 0.02894 | 0.001158 | 11.3 | 0.4522 | utils.py:19(get\_partition\_name) | master |
| 501 | 0.4284 | 0.0008551 | 225.1 | 0.4494 | utils.py:19(get\_partition\_name) | develop (with PR #1022) |
| 28 | 0.03598 | 0.001285 | 13.86 | 0.4951 | utils.py:19(get\_partition\_name)| new changes |

Example of generating a profile:
`python -m cProfile -o tesrun.prof .\scout.py aws -f --profile {profile}`

The important columns are `ncalls`, which is the total number of calls to the function, and `cumtime`, which is the cumulative time spent in the function (and all sub-functions).

The proposed change is to expose the partition through the `AWSFacade` which is available when processing the resources and thus reducing the number of calls and could even avoid hitting the API rate limit.

Fixes issue #935 and PR #1022

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works (optional)
- [ ] New and existing unit tests pass locally with my changes

PR #1341, opened 2021-09-02
Fix ARN processing changes

lowSoA:
Fixed error introduced by #1269 and removed related unused imports.


alessandrogonzalez (approved):
LGTM

PR #1446, opened 2022-08-20
Mitigation for info disclosure vulnerability

### LINKED ISSUE #1436: ScoutSuite does not flag credentials in Lambda Environment Variables
I accidentally discovered credentials in Environment Variables of Lambda functions but SS did not even flag them. In my opinion, this is exactly similar to the finding "Potential Secret in Instance User Data". Keywords like "secret" and "password" should get flagged.

The screenshot below demonstrates an environment variable containing FTP credentials found by SS but not flagged.
ev

Similarly tons of other credentials and critical API keys were found in environment variables. Anyone with reader role can read those credentials.

arparsec:
# Description

Lambda variable value disclosure to user with read permission.

CWE-200: Exposure of Sensitive Information to an Unauthorized Actor
https://cwe.mitre.org/data/definitions/200.html

CWE-668: Exposure of Resource to Wrong Sphere
https://cwe.mitre.org/data/definitions/668.html
        
CWE-664: Improper Control of a Resource Through its Lifetime
https://cwe.mitre.org/data/definitions/664.html

## Type of change

- [ x] Security bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/ScoutSuite/providers/aws/resources/awslambda/functions.py
+++ b/ScoutSuite/providers/aws/resources/awslambda/functions.py
@@ -68,6 +68,14 @@ class Functions(AWSResources):
         # to detect a potential secret and write a constant value instead of the 
         # actual secret value.
         function_dict["env_variables"] = {}
+        function_dict["env_variable_names"] = []
+        function_dict["env_variable_values"] = []
+        redacted_string = "<redacted>"
         if env_variables:
             for env_var_name, env_var_value in env_variables.items():
-                function_dict["env_variables"][env_var_name] = "<redacted>"
+                function_dict["env_variables"][env_var_name] = redacted_string
+                function_dict["env_variable_names"].append(env_var_name)
+                function_dict["env_variable_values"].append(env_var_value)
+
+
+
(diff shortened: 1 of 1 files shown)


michyweb:
Hi @arparsec, could you provide information about what an unauthorized actor may be? This tool is intended to be used by security consultants, pentesters or anybody who wants to discover misconfigurations or for instance secrets. Based on your comments:
we could use something like detect-secrets to detect a potential secret

By relying exclusively on an automated check (via detect-secrets), we would be subjecting the results of the tool to false negatives. Ultimately, it should be the pentester running the tool who manually verifies if the reported variables contain actual secrets or not.

arparsec:
@michyweb Currently, the generated ScoutSuite report doesn't highlight whether lambda secrets exist or not (unlike the EC2 instance metadata feature). We just collect them and record the variable values in the report without any filters or checks. In the current version, you can have a clean account report with 0 findings, which includes sensitive information from lambda environments. Can you explain to me why you think this is a secure default?

arparsec:
Closing, and adding additional features to allow for a user case that wants to explicitly dump all lambda environment info.

michyweb:
This information is collected because there are related checks and findings in NCC Scout, the paid ScoutSuite version

yellow-starburst:
Why is this closed? This is a great feature! 
@michyweb @fernando-gallego

Issue #1465, opened 2022-10-05
[kubernetes] ERROR __main__.py L237: Initialization failure: 'data'


bprsyn:
**Describe the bug**

I run this command on AWS EKS:
```
➜  ScoutSuite git:(develop) scout kubernetes --debug
2022-10-05 14:23:40 bprevost-XPS-13-9300 scout[167757] INFO Launching Scout
2022-10-05 14:23:40 bprevost-XPS-13-9300 scout[167757] INFO Authenticating to cloud provider
2022-10-05 14:23:41 bprevost-XPS-13-9300 scout[167757] INFO GET /version
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/componentstatuses
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/configmaps
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/endpoints
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/events
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/limitranges
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/namespaces
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/nodes
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/persistentvolumeclaims
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/persistentvolumes
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/pods
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/podtemplates
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/replicationcontrollers
2022-10-05 14:23:42 bprevost-XPS-13-9300 scout[167757] INFO GET /api/v1/resourcequotas
[...]

PR #1466, opened 2022-10-05
Adding no data condition on resources

bprsyn:
# Description

Fix for resource without data key.

Explain to #1465


liyun-li:
Merged https://github.com/nccgroup/ScoutSuite/pull/1468 instead. Thanks though.

PR #1515, opened 2023-03-23
Case insensitive conditionals

x4v13r64:
# Description

AWS IAM policy conditions are case insensitive, but SS' condition evaluation is case sensitive. This means that it's raising false positives, e.g. if a resource policy has a `aws:SourceOwner` field whereas SS is looking for `AWS:SourceOwner` ([like here](https://github.com/nccgroup/ScoutSuite/blob/master/ScoutSuite/providers/aws/rules/conditions/policy-statement-poor-condition.json#L17)).

This PR adds case insensitive conditions and uses them  where appropriate. A number of tests are also added to `policy-statement-poor-condition.json`.

## Type of change


liyun-li (commented):
@x4v13r64 - Would it make more sense then to just make `withKey` a case-insensitive match?

PR #1545, opened 2023-06-27
Ensure correct SQL DB AD Admin check

ncc-akis:
# Description

ScoutSuite has a bug when assessing the rule entitled "Azure Active Directory Admin Not Configured for SQL Servers".

As indicated within the CIS benchmark, this should be tested by retrieving the list of AAD Administrators for the database, and failing if the list is empty. The ScoutSuite implementation had two errors:

1. The rule checked for `null` rather than `empty`
2. The facade returned an object which was iterable but not a list, which would not be correctly evaluated.

The fix for the 2nd issue brings the code specific to this check into line with other facade methods that return lists, each of which also use the `list()` constructor to force other iterables into lists at this point. For example, see `sqldatabase.py` line 48.

I have scanned the Azure facade files for similar issues, and fixed one other function `get_database_replication_links()`. This did not cause a bug because the only function which used this converted the result into a `list` by itself. Nonetheless I modified this in order to ensure that all facade functions behave in a consistent way.

## Type of change

## Checklist:

### DIFF
--- a/ScoutSuite/providers/azure/facade/sqldatabase.py
+++ b/ScoutSuite/providers/azure/facade/sqldatabase.py
@@ -56,8 +56,8 @@ class SQLDatabaseFacade:
         try:
             client = self.get_client(subscription_id)
             return await run_concurrently(
-                lambda: client.replication_links.list_by_database(
-                    resource_group_name, server_name, database_name)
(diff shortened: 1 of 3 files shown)


fernando-gallego (approved):
Works for me and doesn't break anything.

PR #1547, opened 2023-06-28
Use Azure ADE status to prevent false positives

ncc-akis:
# Description

In certain circumstances, ScoutSuite can generate false positive reports regarding the rules:

- OS and Data Disks Not Encrypted with CMK
- Unattached Disks Not Encrypted with CMK

This can occur when the VMs are configured to use Azure Disk Encryption (ADE). In this case, Server Side Encryption (SSE) is still enabled, but cannot be used in conjunction with CMK. Because ADE is a suitable alternative which provides stronger protection than SSE, it is not correct to flag the lack of CMK as an issue when ADE is enabled.

The changes in this PR ensure that the presence of ADE is detected. The simplest way I found to do this was to look for the `encryption_settings_collection` attribute in the `raw_disk` object. Based on my testing it seems that this attribute is *only* present when ADE is enabled, and it is set to `None` otherwise. Furthermore, the `encryption_settings_version` property must be set to 1.0 or 1.1 (see https://learn.microsoft.com/en-us/python/api/azure-mgmt-compute/azure.mgmt.compute.v2022_07_02.models.encryptionsettingscollection?view=azure-python). Practically it might be 1.1 only but 1.0 is mentioned in the doc as related to "Azure Disk Encryption with AAD app" so I thought I'd include it.

Because ADE requires the presence of extensions named "AzureDiskEncryption" or "AzureDiskEncryptionForLinux", I decided also to modify the "Virtual Machine Extensions Installed" rule, to prevent it triggering in the presence of those extensions. 

## Type of change

## Checklist:


fernando-gallego (approved):
LGTM.

PR #1593, opened 2023-12-12
Feature/1576 flag permission policies that deny api actions

### LINKED ISSUE #1576: Feature request: flag permission policies that deny access to API actions
**Is your feature request related to a problem? Please describe.**

Is your feature request related to a problem? Please describe.

The presence of Deny statements in permission policies is not necessarily a problem: they have totally valid uses.  However, attempting to control access to API actions in AWS using a wildcard allow and a deny-list is very difficult and subject to problems since there is frequently more than way to accomplish something, new API actions get added all the time, and permission policy authors frequently fail to consider all available API actions.  `s3:GetObject` is denied? Try `s3:GetObjectVersion` instead. `kms:ScheduleKeyDeletion` is denied?  You can still use `kms:PutKeyPolicy` to make the key unusable by everyone.  Etc.  And these aren't hypotheticals: I've seen them while reviewing accounts.

**Describe the solution you'd like**

ScoutSuite should flag permission policies containing Deny statements for specific API actions so that a reviewer can verify that they are appropriate given the account's access control regime and requirements, and that they do not have trivial bypasses.

alexanderdean111:
# Description

Added two new rules to address #1576 

Fixes #1576 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


rdegraaf (changes requested):
You check for Deny statements in inline User permission policies and in Managed permission policies, but not in inline Role or inline Group permission policies.  Also, is there a way to write a single check for all four?  I'm not a fan of the pattern of having four separate checks for everything.

Issue #1596, opened 2024-01-15
GCP Public access buckets are never flagged


fgimenezm:
When scanning a GCP Project with a bucket with public access enabled (AllUsers), the current ScoutSuite logic will *never* flag it.

The current ScoutSuite logic is as follows:
```
    "conditions": [
        "and",
        ["or",
            [
                "cloudstorage.projects.id.buckets.id.member_bindings",
                "withKey",
                "_ARG_0_"
            ],
            [
                "cloudstorage.projects.id.buckets.id.acls",
                "containString",
                "_ARG_0_"
            ]
        ],
        [
            "cloudstorage.projects.id.buckets.id.public_access_prevention",
            "notEqual",
            "enforced"
        ],
        [
            "cloudstorage.projects.id.buckets.id.public_access_prevention",
            "notEqual",
            "inherited"
        ]
    ],
```

but according to Google documentation about [public access prevention](https://cloud.google.com/storage/docs/public-access-prevention)

> the bucket state can only be set to enforced or inherited

so one of these will always be *false*
```
        [
            "cloudstorage.projects.id.buckets.id.public_access_prevention",
            "notEqual",
            "enforced"
        ],
        [
            "cloudstorage.projects.id.buckets.id.public_access_prevention",
            "notEqual",
            "inherited"
        ]
```

making the whole condition always *false*.

Credits to @martinpestoni who found the issue.

x4v13r64:
https://github.com/nccgroup/ScoutSuite/pull/1597 fixes this bug

PR #1597, opened 2024-01-17
Bugfix/GCP Public access buckets are never flagged #1596 

x4v13r64:
# Description

Fixes https://github.com/nccgroup/ScoutSuite/issues/1596 by fetching the effective value of `constraints/storage.publicAccessPrevention` for buckets' projects and evaluating that in the findings.

## Type of change


fernando-gallego:
Hi @x4v13r64 , currently it seems that this check is always flagged, as cloudstorage.projects.id.buckets.id.effective_public_access_prevention always evaluates to False.

PR #1610, opened 2024-02-27
Fixes 'Key Vault Not Recoverable' check

### LINKED ISSUE #1606: False negatives for Azure 'Key Vault Not Recoverable' rule
**Describe the bug**

**ScoutSuite fails to flag unrecoverable key vaults as such, if the Azure API returns `enable_soft_delete = null` for the key vault.**

ScoutSuite uses the computed property `recovery_protection_enabled` to flag vaults where soft-delete is disabled or purge protection is missing:

```python
        vault[
            'recovery_protection_enabled'] = raw_vault.properties.enable_soft_delete and \
                                             bool(raw_vault.properties.enable_purge_protection)
```

However, the code does not correctly handle all possible API return values. According to my tests, `enable_soft_delete` can be `true`, `false`, or `null`.

**I have verified in the Azure Portal that soft-delete is disabled for vaults with `enable_soft_delete = null`**

In the case of a `null` API response, the code sets `vault['recovery_protection_enabled'] = None`.

The `keyvault-not-recoverable.json` rule however checks only for `false` to flag affected resources. The rule does not handle `None` / `null`: 

```python
    "conditions": [
        "and",
        [
            "keyvault.subscriptions.id.vaults.id.recovery_protection_enabled",
            "false",
            ""
        ]
    ],
```

**This results in false negatives**. Some Key Vaults lacking soft-delete (and, by extension, purge protection) are not flagged.

**To Reproduce**

I was not able to reproduce this issue for recently created vaults, for which either `true` or `false` was returned by the API. I believe the API returns `null` for old key vaults that were created before soft-delete was added and that have not been migrated yet.

The issue can therefore likely only be reproduced in older tenants that have Key Vaults that are a few years old and that have not enabled soft-delete yet.

1. Run ScoutSuite with default settings

```bash
scout azure --cli --subscriptions '<subscription-id>'
```

2. Search for affected Key Vaults using `jq`:

```bash
jq -r """.services.keyvault.subscriptions[].vaults[]                                                                                                         
| select(             
    (.properties.enable_soft_delete == null)
  )
| .name""" < scoutsuite-results.json
```

rieck-srlabs:
# Description

ScoutSuite *incorrectly did not* flag key vaults for which the API returned `enable_soft_delete = null`. Such key vaults have neither soft-delete nor purge protection enabled and are also not recoverable. The check would only flag key vaults with `enable_soft_delete = false`

Fixes #1606 

## Type of change

## Checklist:

- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/ScoutSuite/providers/azure/resources/keyvault/vaults.py
+++ b/ScoutSuite/providers/azure/resources/keyvault/vaults.py
@@ -30,7 +30,7 @@ class Vaults(AzureResources):
         vault['resource_group_name'] = get_resource_group_name(raw_vault.id)
         vault['properties'] = raw_vault.properties
         vault[
-            'recovery_protection_enabled'] = raw_vault.properties.enable_soft_delete and \
+            'recovery_protection_enabled'] = bool(raw_vault.properties.enable_soft_delete) and \
                                              bool(raw_vault.properties.enable_purge_protection)
         vault['public_access_allowed'] = self._is_public_access_allowed(raw_vault)
         vault['rbac_authorization_enabled'] = raw_vault.properties.enable_rbac_authorization


fernando-gallego:
LGTM.

PR #1723, opened 2026-03-13
Fix false positive root MFA when org centralized root access is enabled

### LINKED ISSUE #1710: False positive for AWS "Root Account without MFA"
**Describe the bug**

Please provide:
ScoutSuite gives false positive result for "Root Account without MFA" if you are using AWS Organization and have disabled root access completely by the best security practices from AWS via ["Centrally managing root access for customers using AWS Organizations"](https://aws.amazon.com/blogs/aws/centrally-managing-root-access-for-customers-using-aws-organizations/)

**To Reproduce**
1. Create AWS organization
2. Add child account
3. Disable root access completely following this AWS guide - https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-enable-root-access.html
4. Run scoutsuite agains a child account.

Please provide:

- The **exact** CLI parameters used to run Scout Suite.
- Any specific configuration within the cloud account which might have lead to the issue.

Nothing specific here, just: `python scout.py aws -r <region>`

raajheshkannaa:
## Summary

Fixes #1710

When AWS Organizations centralized root access management is enabled and root credentials are removed from member accounts, ScoutSuite incorrectly flags "Root Account without MFA". This happens because the credential report shows `mfa_active` as `false` for the root user, but MFA cannot and need not be configured when root credentials have been centrally removed.

**Changes:**

- Added `get_organizations_root_credentials_managed()` to the IAM facade, which calls `iam:ListOrganizationsFeatures` to check if `RootCredentialsManagement` is enabled. Fails gracefully (returns `False`) when the account is not in an org or lacks the required permission.
- In `credentialreports.py`, the result is used to set a `root_credentials_managed_centrally` flag on the root credential report entry.
- Both `iam-root-account-no-mfa.json` and `iam-root-account-no-hardware-mfa.json` rules now include a condition to skip accounts where this flag is `true`.
- Updated test fixture `iam-root.json` with the new field and a test case for the centralized root access scenario.

## Test plan

- [ ] Run ScoutSuite against an AWS account **without** Organizations or centralized root access. Verify root MFA finding still fires when MFA is not enabled.
- [ ] Run ScoutSuite against an AWS Organizations member account **with** centralized root access enabled and root credentials removed. Verify root MFA finding does NOT fire.
- [ ] Run ScoutSuite against an account where `iam:ListOrganizationsFeatures` is not permitted. Verify the API call fails silently and root MFA finding behaves as before (no regression).
- [ ] Run `python -m pytest tests/test_rules_processingengine.py` to validate the rule engine test with updated fixtures.

### DIFF
--- a/ScoutSuite/providers/aws/facade/iam.py
+++ b/ScoutSuite/providers/aws/facade/iam.py
@@ -67,8 +67,9 @@ class IAMFacade(AWSBaseFacade):
             response = await run_concurrently(client.list_organizations_features)
             enabled_features = response.get('EnabledFeatures', [])
             return 'RootCredentialsManagement' in enabled_features
-        except Exception:
+        except Exception as e:
             # Expected to fail when not in an org or lacking iam:ListOrganizationsFeatures permission
+            print_warning(f'Could not check Organizations root credentials management: {e}')
             return False
 
     async def get_groups(self):
--- a/ScoutSuite/providers/aws/facade/iam.py
+++ b/ScoutSuite/providers/aws/facade/iam.py
@@ -56,6 +56,21 @@ class IAMFacade(AWSBaseFacade):
                 print_exception(f'Failed to download credential report: {e}')
             return []
 
+    async def get_organizations_root_credentials_managed(self):
+        """
+        Check if centralized root access management is enabled for the organization.
+        Returns True if RootCredentialsManagement is among the enabled features.
+        Returns False if the API call fails (e.g. not in an org, insufficient permissions).
+        """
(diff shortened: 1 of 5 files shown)


raajheshkannaa:
Retargeted this from `master` to `develop` per CONTRIBUTING step 6. Fix and feature merges land on `develop`, `master` only takes release merges, so this was sitting on the wrong base.

The diff is unchanged (5 files) and it merges cleanly against `develop`. Open since March with no review, happy to rebase or split it if that makes it easier to look at.

